PR #16, opened 2021-06-24
Add drop and alter DDL and expand create to views

dmfay:
What do you think about renaming the `table_reference` node to `object_reference` at some point, since the format is identical for views (& functions too, for create/alter/drop)?


DerekStride (approved):
> What do you think about renaming the table_reference node to object_reference at some point, since the format is identical for views (& functions too, for create/alter/drop)?

Yeah I think that could make sense, I like how `table_reference` is more specific though, although I'm not tied to that idea.

DerekStride:
We should update the feature list on the Readme too

PR #19, opened 2021-08-14
Arbitrary function calls + Node bindings

dmfay:


DerekStride (approved):
Awesome work as always @dmfay, thanks!

Issue #18, opened 2021-07-06
Other parser implementations


m-novikov:
I didn't see your project when I started my parser implementation https://github.com/m-novikov/tree-sitter-sql
I added you to the readme, but maybe we should join forces somehow :)

dmfay:
There's also [tjdevries/tree-sitter-sql](https://github.com/tjdevries/tree-sitter-sql) and [tk-shirasaka/tree-sitter-sql](https://github.com/tk-shirasaka/tree-sitter-sql) that I know of.

DerekStride:
> On future plans I do think it'd be good to talk with the maintainers of other tree-sitter-sql projects -- maybe we could set up a chatroom of some sort so everyone doesn't have to coordinate right off the bat? There's still no official binding and it'd be nice to consolidate our various efforts and get that process started.

Great idea!

Also, sorry for the late reply here @m-novikov, I'd be happy to add a list of other implementations. I think the suggestion from @dmfay to consolidate our efforts makes a lot of sense. I'm open to suggestions and ideas on how to do this. In the meantime I can set up a chatroom (e.g. discord, slack, etc) add a link to the readme. I'll also invite other maintainers to join the conversation.

If the chatroom is desired leave a 👍 on this comment and I'll set that up.

FFengIll:
I am sure tree-sitter is good.
But for sql, the kinds of dialects could be a large challenge.

I wanna provide a ref for help (more or less) : https://github.com/AlecStrong/sql-psi
Look forward to the solution with tree-sitter ~

aharpervc:
What's a good strategy to have multiple SQL variants live side by side in a single repo... and is such a thing even realistic? For example, maybe it's better that pg sql & sqlite are distinct grammars/repos.

I'm also think about this from the perspective of making an mssql grammar. I don't necessarily think it'd be useful to open a PR for that to any of these repos. Perhaps a forking to tree-sitter-mssql and making the adjustments there would be preferable.

dmfay:
> What's a good strategy to have multiple SQL variants live side by side in a single repo... and is such a thing even realistic? For example, maybe it's better that pg sql & sqlite are distinct grammars/repos.

I can't speak for anyone else but I don't think it's unrealistic as such. It's a good question and might not have an easy answer.

[...]

PR #20, opened 2021-09-01
another batch

dmfay:
It may be easiest to look commit by commit, it's a bit of a mess collapsed into one diff.


DerekStride (approved):
I reviewed each commit and overall it looks great. I think the general form of the AST is in pretty good shape. Sorry for the delayed review I've been on vacation for the past 2 weeks. I've been fairly busy in general these last few months but I have a few ideas to make the AST more robust that I hope to capture in an issue soon.

@dmfay, you have been doing great work on this project, would you like to be added as a maintainer so you don't have to wait for my reviews to get PRs merged?

dmfay:
Absolutely! On future plans I do think it'd be good to talk with the maintainers of other tree-sitter-sql projects (#18) -- maybe we could set up a chatroom of some sort so everyone doesn't have to coordinate right off the bat? There's still [no official binding](https://tree-sitter.github.io/tree-sitter/#language-bindings) and it'd be nice to consolidate our various efforts and get that process started.

PR #23, opened 2022-02-18
untangle predicates from assignments/expressions and use the latter consistently

dmfay:
tree-sitter-sql uses predicates (comparison statements with =, >, etc) and binary operations (addition, multiplication) interchangeably, and additionally reuses the predicate rule for assignments in UPDATE statements. These two commits clean that up with minimal effect on output trees:

* assignments are now their own rule
* binary_operations are split out from predicates to become one possible (recursive) choice of _expression
* _expressions now include lists, cases, and function calls, which allows the _expression rule to supplant _field and _function_param

The most noticeable side effect in output is that the optional parentheses in `distinct (col)` now counts the expression as a list node. `returning` clauses are a little different too thanks to the reused select_expression rule.


dmfay:
going ahead with this -- it's a big change in absolute terms but shouldn't be too disruptive!

DerekStride:
Looks great! I've been a little busy with some personal things lately so I didn't get to this sooner. The tests you added and changed looked good to me though.

Issue #24, opened 2022-07-27
Feature Request: Support for parameter placeholders


wilriker:
Hi!

In my daily work I have to deal with literal SQL queries that contain parameter placeholders in 99% of the cases.
Is it possible/planned/not planned to add support for them? The drivers I use expect `?` as placeholder although there are other syntaxes.

Or is this something that can possibly be even added at the level of queries?

Cheers
Manuel

DerekStride:
An example would be very helpful and you looking for something like parameterized queries?

```sql
SELECT *
FROM x
WHERE id = ?;
```

wilriker:
> ```sql
> SELECT *
> FROM x
> WHERE id = ?;
> ```

That's exactly what I meant. 👍

DerekStride:
Alright, it should be working now 🎉

wilriker:
Thanks! I can confirm that `?` is now recognized as parameter!

PR #26, opened 2022-07-27
Parameterized Query support

DerekStride:
## What

Addresses https://github.com/DerekStride/tree-sitter-sql/issues/24

The grammar currently doesn't support parameterized queries / prepared statements and displays an error node when the parser encounters a `?` token. I've added a dedicated `parameter` node to the list of valid expressions.

cc// @wilriker


grammar.js
```diff
@@ -783,6 +783,8 @@ module.exports = grammar({
       '*',
     ),
 
+    parameter: $ => "?",
```
dmfay, line 786: The `$1` syntax is also quite common and should be straightforward to capture here if you're so inclined.

dmfay (approved):

Issue #25, opened 2022-07-27
Bug: support SELECT from more than two tables


wilriker:
As of right now a select like
```sql
SELECT *
FROM 
    Table1 a,
    Table2 b,
    ...
    TableN z
WHERE...
```
will result in an error on the table list. It does recognize both the first and the last table with its `name` and `table_alias` just fine but it does not recognize the table(s) in between.

DerekStride:
This should be supported now, reopen the issue or create a new one if you have any trouble with it

wilriker:
I can confirm it works as expected now. Thanks!

PR #27, opened 2022-07-27
Add support for selecting from multiple tables

DerekStride:
# What

Addresses https://github.com/DerekStride/tree-sitter-sql/issues/25

We currently only accept a single table table expression in the `from` clause. I've added an anonymous node to allow multiple `table_expression` nodes in succession.

I've only changed the `from` node to use this the new path though. I've left `_delete_from` & `join` using the regular path of only allowing a single `table_expression`, it's be an easy update to use it everywhere if thats valid.

cc// @wilriker


dmfay (approved):
I don't think the old-style cartesian product join syntax is legal in joins or deletes so no worries there. Update-from is a maybe, I'll have to check, but we don't support that yet anyway.

PR #29, opened 2022-08-03
Postgresql cast syntax

### LINKED ISSUE #28: Dbt Jinja
Hi there, 
I am using this Treesitter grammar for a while now in neovim. I work a lot with dbt, which consists of sql files with some jinja blocks inside. Is there any change to get this included in the grammar (or at least ignored).
I have some (pg)SQL files which look really weird, e.g. the Syntax Highlighting stops in the middle of the file and TS reports and error (even without jinja blocks near the error location).

DerekStride:
## What

Fixes #28 

This PR does 2 things:

1. Adds an alias node which makes connecting the value and aliased name nodes easier in the AST.
2. Adds a cast node that represents postgresql's `value::type` syntax.

The first commit has all the changes related to the new alias node and updates the tests affected. The second commit adds the new cast node.

cc\\ @matthias-Q


dmfay:
I actually tried tackling this + function-style casting yesterday! I ran into some ambiguity issues with timestamp types of all things but I think I've put that to rest -- I'll run that up and we can compare notes. My version's a little more invasive atm, since I removed the sort-of-redundant `function_call` in the process.

It does look like your deanonymized `alias` is more of an `aliased_field` now? And since it's an `_expression` it would be assumed valid now wherever `_expression` is used, including in `ORDER` clauses or inside predicates, where it isn't legal.

PR #30, opened 2022-08-03
Casting with function-style cast() and Postgresish field::type

dmfay:
#29 , alternate version


test/corpus/casting.txt
```diff
@@ -0,0 +1,46 @@
...
+  (select
+    (keyword_select)
+    (select_expression
+      (cast (identifier) (literal) (keyword_as) (keyword_text))))))
+
+==================
+Aliased
+==================
+
+SELECT CAST(1 AS TEXT) AS fieldname;
```
DerekStride, line 20: Not a blocker for this PR but I do think we'll need a node to enclose an aliased field like this eventually. We can wait for a use case though.

DerekStride (approved):
This looks great, much better than my PR imo 👍

PR #31, opened 2022-08-24
Allow lazy group bys

matthias-Q:
The following statement results currently in an error:
```sql
SELECT 
a, 
SUM(b) 
FROM tab 
GROUP BY 1
```

I think that this is an SQL anti pattern, however it is commonly used by many users, thus the parsing breaks in many cases.

I am also not sure if the edits I made on the regex could be improved. Could you please have a look at my changes.

Thanks, 
Matt

P.S. Sorry for the formatting of this PR. This is one of my first contributions to open source projects.


dmfay:
welcome! I use this syntax myself often enough, imo it's fine as long as it doesn't go to production :grinning: 

`identifier` isn't quite the right place to be making this change -- what you've done allows _any_ identifier to start with a number, which isn't valid in SQL (i.e. you can't name a table `2_things`). But the `group by` column index shortcuts aren't identifiers, it's something you can put in the place _of_ an identifier. You probably saw the `group_by` definition at line 1037; we need to make that offer a `choice` between `identifier` and `_number`, which is defined just above the former.

(we also need to make it support lists of grouping targets, it seems, but we can do that later)

matthias-Q:
Ah yes, I was suspecting that this might be an issue.
So, I have tried the following for the group by definition instead:

```js
    group_by: $ => seq(
      $.keyword_group_by,
      choice(
          $.identifier,
          $._number,
      ),
      optional($._having),
    ),
```

However, the tests fails now because the AST looks like this
```
(program [1, 0] - [6, 0]
  (statement [1, 0] - [4, 21]
    (select [1, 0] - [1, 28]
      (keyword_select [1, 0] - [1, 6])
      (select_expression [1, 7] - [1, 28]
        (field [1, 7] - [1, 15]
          name: (identifier [1, 7] - [1, 15]))
        (count [1, 17] - [1, 28]
          name: (identifier [1, 17] - [1, 22])
          parameter: (field [1, 23] - [1, 27]
            name: (identifier [1, 23] - [1, 27])))))
    (from [2, 0] - [4, 20]
      (keyword_from [2, 0] - [2, 4])
      (table_expression [2, 5] - [2, 13]
        name: (identifier [2, 5] - [2, 13]))
      (group_by [3, 0] - [4, 20]
        (keyword_group_by [3, 0] - [3, 8])
        (keyword_having [4, 0] - [4, 6])
        (predicate [4, 7] - [4, 20]
          left: (field [4, 7] - [4, 15]
            name: (identifier [4, 7] - [4, 15]))
          right: (literal [4, 18] - [4, 20]))))))
```

After the keyword_group_by, there is no neither in identifier nor a number, which is odd.

dmfay:
Syntax tree elements starting with an underscore are anonymous (there's an `_identifier` to match `identifier` too). I don't think we need a named `number` for anything else so instead we can `alias($._number, $.literal)` like we do in parametric types to give you a `literal` node after the `keyword_group_by`.

matthias-Q:
Wouldn't it be better to give this new node a uniform name? Now it is either an `identifier` or a `literal`. When someone wants to query these elements, there are now 2 different possibilities.

dmfay:
It looks like you can in fact group by any `_expression`, arithmetic, subqueries, the works, some of it like binary expressions semantically meaningless but valid SQL, at least in Postgres. It'll even parse `distinct`, which presumably doesn't do much either but hints that parser treats it like a select list.

`_expression` allows even more types under an anonymous node which doesn't help what you're talking about, but a prepared statement parameter isn't a function call and so on; we _need_ those to be labeled what they are. I think the best thing to do is to add a more queryable container for just the group expressions (and fix the list case in the process), like how `select_expression` and `_select_expression` work for select lists, minus `all_fields` and aliasing which aren't allowed in `group by`. Your final syntax tree would look like this:

```lisp
(...
 (group_by
  (group_list
   (literal)
   (field ...))))
```

matthias-Q:
So, I added the `group_list`. 
I was not able to find an example that subqueries, arithmetic expressions, binary expressions... are valid SQL syntax. So I have excluded that for now. I hope that's ok.

dmfay (approved):
This looks great, thanks!

PR #32, opened 2022-08-26
Window functions

matthias-Q:
I implemented nodes for window_functions based on:

https://en.wikipedia.org/wiki/Window_function_(SQL)#Syntax
and 
https://github.com/postgres/postgres/blob/master/src/backend/parser/gram.y (Line 15829)

- [x] `PARTITION BY`
- [x] `ORDER BY`
- [x] `frame_clause`
- [x] `PARTITION BY x ORDER BY y`
- [x] `PARTITION BY x ORDER BY y frame_clause`
- [x] reusable window functions

I am a bit unsure about anonymous nodes and if window functions can have multidimensional partitioning/ordering.

Window functions can also be defined at another location and then just referenced after the `OVER` keyword. Not sure if this case can be covered at all.


matthias-Q:
Reusable window definitions (`$.window_clause`) are now parse-able. However, I am not sure if I referenced it at the right location (line 307).

Since I had to add a lot of keywords, do you think that we need more unit test cases? There seems to be many different ways to construct window functions.

grammar.js
```diff
@@ -905,6 +925,99 @@ module.exports = grammar({
...
+        choice(
+            $.keyword_range,
+            $.keyword_rows,
+            $.keyword_groups,
+        ),
+        optional(
+            $.keyword_between,
+        ),
+        $._frame_definition,
+        $.keyword_and,
```
dmfay, line 981: `and` is only required with the `between`, which to be strictly correct would probably require a choice() of seq(between, _frame_definition, and, _frame_definition) or seq(_frame_definition, _frame_definition).

grammar.js
```diff
@@ -285,6 +304,7 @@ module.exports = grammar({
         $._insert_statement,
         $._update_statement,
       ),
+      optional($.window_clause),
```
dmfay, line 307: I think this is the right place, since iirc window functions can be used with updates at least

grammar.js
```diff
@@ -905,6 +925,99 @@ module.exports = grammar({
...
+        $.keyword_window,
+        $.identifier,
+        $.keyword_as,
+        $.window_specification,
+    ),
+
+    window_specification: $ => seq(
+        '(',
+        repeat(
+            choice(
```
dmfay, line 1003: at least with Postgres  you get up to one of each of `partition by`, `order by`, and frame clause in that order. MySQL docs appear to confirm ordering, and I don't think we need to worry about multidimensional windowing, so this can be a seq() of optional()s.

grammar.js
```diff
@@ -905,6 +925,99 @@ module.exports = grammar({
       ')',
     ),
 
+    window_partition_by_expression: $ => seq(
+        $.keyword_partition_by,
+        $.identifier,
+    ),  
+
+    window_order_by_expression: $ => seq(
```
dmfay, line 933: `using` and `nulls first/last` are allowed in statement `order by` clauses as well, so it'd be good to update that and reuse it instead of creating a near clone!

grammar.js
```diff
@@ -905,6 +925,99 @@ module.exports = grammar({
       ')',
     ),
 
+    window_partition_by_expression: $ => seq(
```
dmfay, line 928: I think `partition_by` is fine here, especially if we can reuse `order_by` as mentioned below

grammar.js
```diff
@@ -905,6 +925,99 @@ module.exports = grammar({
...
+            $.keyword_current_row,
+            seq(
+                alias($._number, $.literal),
+                $.keyword_following,
+            ),
+            $.keyword_unbounded_following,
+        ),
+    ),
+
+    window_frame_clause: $ => seq(
```
dmfay, line 971: `window_frame` suffices here imo

dmfay (commented):
We're going to need even more keywords to account for the parser's space sensitivity -- `unbounded preceding` as a single keyword fails to recognize `unbounded    preceding`. At least some are reusable like `group`.

Solid choices with anonymous nodes. `_frame_definition` is the only one I'm thinking twice about since I can see wanting to target those specifically, and right now it's dropping literals and keywords into the `window_frame_clause`.

I think this is a decent amount of tests to start with, but I do have a few suggestions to add:

* empty window
* two inlined windows
* one inlined and one named window

matthias-Q:
Thanks for the comments. I have implemented your suggestions.

dmfay:
there are still some multi-word keywords that need to be broken up: `partition by`, `current row`, and the excludes. SQL doesn't care about the number and nature of spaces (tabs, newlines, etc) between them but tree-sitter considers a single-keyword `partition by` to contain exactly one space, and will fail to match any other configuration. (I didn't realize Github would condense the spaces in my example.... helpful!)

Everything else looks good though!

matthias-Q:
Now I get it! I was looking at the `order by` keyword, which was implemented a while ago. I will also break this up and create anonymous nodes like `if not exists` for these keywords.

dmfay (approved):

PR #33, opened 2022-08-29
Generalize lists after group by and partition by

matthias-Q:
I checked and you can have multiple columns after a `partition by`, similar as after `group by`. Thus, I tried to generalize this as `expression_list`.

However, I think this can be generalized further. Is that similar to `order by` where we use `_expression`?


matthias-Q:
I noticed that subqueries and the complete field `schema.table.column` are also valid in `partition` and `group by` lists.

dmfay:
we can't generalize it to the `order by` case because of directions, `using`, and `nulls first/last`. Good work on this though! No notes :)

I did find an issue with `order_expression` itself not getting all available directives on subsequent columns when I looked though so I'll push another pr up separately -- there's one point I'm not 100% sure on.

PR #34, opened 2022-08-30
Fill out operators for order by using and make using/nulls clauses av…

dmfay:
…ailable to subsequent order targets

@matthias-Q question here is: do we wrap each field-plus-clauses in an `order_target` (or something) node? I'm leaning yes for specificity's sake, but only leaning -- any thoughts?


matthias-Q:
A `order_target` node sounds like a good idea.

PR #35, opened 2022-08-30
Missing Possibilities in Group and Partition Expression lists

matthias-Q:
Hi,
this seems to be an absolute rabbit hole. I figured out that almost everything from the `_expression` node is valid in the `partition by` and `group by` list. 

Exceptions are:
- `window functions`
- `count` (at least I was not able to write a valid statement)
- `list` (because it is self referencing  the `_expression` node)

I did not write tests yet, because I want to make sure that we don't miss things here this time. It might be a good idea to put these tests (together with the other window functions tests) in a separate file.


dmfay:
yeah, you can partition and group (and order!) by all manner of things, most of which don't often make much semantic sense. I don't think we need to be super thorough about testing each and every permutation here; one big test case with a single partition by a, b, c, d, ..../order by same is about the maximum I'd bother with! Agreed it'd be good to break the window functions out of the now-2300-line select corpus though :)

matthias-Q:
I added a very big example in the new window_functions corpus. 
I re-based main into my branch, but it looks like there are still merge conflicts. Please let me know if I should do something in order to resolve them. I mean, the parser.c just need to be freshly generated, right?

matthias-Q:
One more thought: isn't `order_target` an extension of `expression_list`?

dmfay:
grammar.js is the only file to be careful with, yes. It looks like your rebase wasn't quite correct since it sandwiched my commits between yours instead of fast-forwarding mine and applying yours after. If you don't want to try to repair it (at a guess, interactive rebase to delete my commits, pull DerekStride/main into your main, and try rebasing again) you could cherry-pick your commits into a freshly cut branch, delete the old branch, change the name of the new one, and force-push -- either way you'll need to regenerate and commit the parser files, but the pull request should pick it up.

> One more thought: isn't order_expression an extension of expression_list?

It's very close! But order directionality and the other context-specific clauses mean they aren't interchangeable. Say we combined them: now when we're parsing an `expression_list` and run into a `desc` after we finished the immediate child `field` or `case`, how do we know it's supposed to be here? We're out on a leaf of the syntax tree and don't have a way to express "this is only valid if my grandparent is an `order_by`".

matthias-Q:
Thanks, is was able to fix the issues.

OK, I guess you are right and it not possible to have ´order_expression´  and ´expression_list´ as a same node type. Thanks for the explanation!

PR #36, opened 2022-09-02
From Clauses are missing subqueries

matthias-Q:
Currently the grammar only supports to select from a table (`table_expression`) as a list of tables with optional schema prefix.

Sub queries, function call or parameters where not parsed correctly. I have added these choices to `table_expression`. Again, I am not so convinced about the name for the node, since this PR extends it beyond just "tables" (1).

I also think the the wrong node is anonymized. The node in line 1054 should imo be public. I can imagine use cases of directly wanting to select the range of `table_expressions` (2).

There are also (dialect specific?) keywords before the `table_expressions` like `lateral` or `only` (at least in PostgreSQL) (3).

I will add tests, when we clarified these three points.


dmfay:
1. good thought; `table_expression` is probably more properly one choice under a `relation` ^1, to include views, set-returning function results, subqueries, and aliased `values` clauses (`select * from (values (1, 2), (3, 4)) as v;`; optionally `as v (col1, col2)`). It _doesn't_ include `parameters` normally (i.e. no `select * from ?`); can you give an example of what you're thinking of there?
2. I'm okay with deanonymizing this and calling it `relation_list` or the like, but individual relations are also at least as useful targets so we shouldn't anonymize that.
3. `lateral` is a join type which we already support. I'm not sure offhand how it's used in the old-school `from rel, rel, rel` syntax but Postgres does raise a syntax error _after_ the keyword and relation name. `only` is valid before each table name in such a list so I think the call there is to make it optional in the "table" choice path of `relation`.

[1] we do need to keep a separate `table_expression` for `[schema.]name` specifically, since inserts, updates, and deletes can only target tables/views. Aliasing should move into `relation` though since that's more general.

matthias-Q:
I have added a `relation` and `relation_list` node as refactored the `_table_expression` node.

I am not sure how to allow selects from `values` statements (your example: `select * from (values (1, 2), (3, 4)) as v;`).

This is what I tried:

```javascript
_inplace_insert: $ => seq(
    '(',
      $.keyword_values,
      $.list,
    ')',
),
```
However, it's definitely wrong. 


I was mistaken about the parametrization, thus there is no need to add that.

matthias-Q:
Found my error. Now you example also works.

grammar.js
```diff
@@ -1065,12 +1066,66 @@ module.exports = grammar({
...
+              ),
+            ),
+          ),
+          optional(
+            seq(
+              $.keyword_as,
+              field('table_alias', $.identifier),
+              $.list,
+            ),
+          ),
```
dmfay, line 1114: you can specify a list of attributes after any relation (e.g. `select * from now() as n (blah)`) so this can be folded into the `relation` node. Also check out line 690 in this branch for `insert_expression`:

```
      optional(alias($._column_list_without_order, $.column_list)),
```

`list` is too generic, we allow column names and only column names here.

grammar.js
```diff
@@ -1065,12 +1066,66 @@ module.exports = grammar({
...
+      optional(
+        seq(
+          field('schema', $.identifier),
+          '.',
+        ),
+      ),
+      field('name', $.identifier),
+    ),
+
+    _inplace_insert: $ => seq(
```
dmfay, line 1096: this could be reusable for `insert_expression` as a `values` node (also a better name, imo) if `relation` offered a choice `seq('(', $.values, ')'`

dmfay (commented):
oof, the deanonymized `relation_list` is pretty universal in that version, even in the probably-most-common circumstance of querying `from` a single relation and joining others 😬

What do you think of this version? It avoids a single-child `relation_list` without getting weird with precedence:

```
    from: $ => seq(
      $.keyword_from,
      choice(
        $.relation,
        $.relation_list,
      ),
      optional($.index_hint),
      repeat(
        choice(
          $.join,
          $.lateral_join,
        ),
      ),
      optional($.where),
      optional($.group_by),
      optional($.order_by),
      optional($.limit),
    ),

    relation_list: $ => seq(
      $.relation,
      repeat1(
        seq(
          ',',
          $.relation,
        ),
      ),
    ),
```

matthias-Q:
I am not entirely sure, if I understand your last two comments.

Renaming `_inplace_insert` into `values` sounds good to me - and reusing it in the `insert_expression` as well. However, in `insert_expression` there is a `list` after the `keyword_values`. That's why I used that as well.

I tried the following (replacing the `list` with the alias of the column list `alias($._column_list_without_order, $.column_list)`. However with this the test "Select from inplace values insert" fails.

```javascript
values: $ => seq(
    '(',
      $.keyword_values,
      alias($._column_list_without_order, $.column_list),
      optional(
          repeat(
          seq(
            ',',
            alias($._column_list_without_order, $.column_list),
          ),
        ),
      ),
      optional(
        seq(
          $.keyword_as,
          field('table_alias', $.identifier),
          alias($._column_list_without_order, $.column_list),
        ),
      ),
    ')',
),

relation: $ => seq(
  choice(
    $.subquery,
    $.invocation,
    alias($._table_expression, $.table_expression),
    $.values,
  ),
  optional(
    seq(
      optional($.keyword_as),
      field('table_alias', $.identifier),
    ),
  ),
),
```

Replacing the `relation_list` with the `choice` is very good (despite the fact that I had to touche 60+ tests again :-))

matthias-Q:
```javascript
    values: $ => seq(
        '(',
          $.keyword_values,
          $.list,
          optional(
              repeat(
              seq(
                ',',
                $.list,
              ),
            ),
          ),
          optional(
            seq(
              $.keyword_as,
              field('table_alias', $.identifier),
              optional(alias($._column_list_without_order, $.column_list)),
            ),
          ),
        ')',
    ),
```

This would work for the `values` node and I think it is correct, since the first 2 `list`s definitely need to be like this. They cannot be a `column_list`

dmfay:
> Renaming _inplace_insert into values sounds good to me - and reusing it in the insert_expression as well. However, in insert_expression there is a list after the keyword_values. That's why I used that as well.
>
> I tried the following (replacing the list with the alias of the column list alias($._column_list_without_order, $.column_list). However with this the test "Select from inplace values insert" fails.
> (snip)
> This would work for the values node and I think it is correct, since the first 2 lists definitely need to be like this. They cannot be a column_list

I tried describing it but this is pretty fiddly, so here's some code instead based on your last implementation:

```js
values: $ => seq(
  $.keyword_values,
  $.list,
  optional(
      repeat(
      seq(
        ',',
        $.list,
      ),
    ),
  ),
),

relation: $ => seq(
  choice(
    $.subquery,
    $.invocation,
    alias($._table_expression, $.table_expression),
    seq(
      '(',
      $.values,
      ')'
    ),
  ),
  optional(
    seq(
      optional($.keyword_as),
      field('table_alias', $.identifier),
      optional(alias($._column_list_without_order, $.column_list)),
    ),
  ),
),
```

* table _and column_ aliases apply to `relation`s generally, not just to inline `values` clauses (as in my example with a function call), so we can do that all in `relation` instead of adding redundant aliasing to `values` as well
* the parenthesis literals only apply if the `values` clause is being treated as a `relation` (it isn't `insert into x *(* values ('y') *)*`) -- compare to `subquery`, which also brings its own parentheses

> Replacing the relation_list with the choice is very good (despite the fact that I had to touche 60+ tests again :-))

yeah, sorry about that! thanks for being a good sport about it 😅

matthias-Q:
Thanks, yeah it becomes more and more complex :smile:, but I think it looks really good now.

The last test had invalid SQL syntax - the grammar was correct. This took me a while to understand.

dmfay:
merged! There's a little refactoring possible between table_expression and table_reference to do and `only` is also applicable to updates and deletes -- I'll see if I can get that done this evening.

PR #37, opened 2022-09-08
Add numeric data type extensions

matthias-Q:
Postgres has some numeric datatype extensions. For example `smallint` and `int2` yield the same data type.

This PR adds:
* `int2`, `int4` and `int8`
* `serial2`, `serial4` and `serial8`
* `float(p)`

The value p for the precision can be between 1 and 53. To check this, we could use the following regex: `/\\b(0?[1-9]|[1-9][0-9]|100)\\b/`
However, I am not sure if we should and secondly I do not know how to modify the function `parametric_type`.


DerekStride (approved):
> The value p for the precision can be between 1 and 53. To check this, we could use the following regex: /\\b(0?[1-9]|[1-9][0-9]|100)\\b/
However, I am not sure if we should and secondly I do not know how to modify the function parametric_type.

I don't think we need to be restrictive here, it's possible that could change in the future and other tools can use a query to check for that if they need to.

PR #40, opened 2022-09-12
Like & Similiar To Pattern Matching

matthias-Q:
I added pattern matching in where clauses with `like`, `ilike`, `similiar to` and the inverse counterparts.

The resulting tree in the test looks odd. Is it correct, that the predicates are nested?


grammar.js
```diff
@@ -1334,6 +1340,10 @@ module.exports = grammar({
         ['!=', 'binary_relation'],
         ['>=', 'binary_relation'],
         ['>', 'binary_relation'],
+        [$.keyword_like, 'pattern_matching'],
+        [seq($.keyword_not, $.keyword_like), 'pattern_matching'],
+        [seq($.keyword_similar, $.keyword_to), 'pattern_matching'],
```
DerekStride, line 1345: You can use your anonymous nodes here:

```suggestion
        [$._similar_to, 'pattern_matching'],
```

test/corpus/select.txt
```diff
@@ -2016,3 +2016,66 @@ FROM
...
+              (predicate
+                (predicate
+                  (field
+                    (identifier))
+                  (keyword_like)
+                  (predicate
+                    (literal)
+                    (keyword_and)
+                    (field
+                      (identifier))))
```
DerekStride, line 2063: Adding the node names will help make this test clearer, as it stands right now I don't think it's correct but it's hard to tell:

```suggestion
                  (predicate
                    left: (literal)
                    operator: (keyword_and)
                    right: (field (identifier))))
```

grammar.js
matthias-Q: Done

test/corpus/select.txt
matthias-Q: I added the verbose node indetifiers

matthias-Q:
I think it is so confusing, because the `keyword_and` is also an operator and
```
expression1 and expression2
```
follows the same pattern of `left operator right`. This results in the nesting.

grammar.js
```diff
@@ -1334,6 +1340,10 @@ module.exports = grammar({
         ['!=', 'binary_relation'],
         ['>=', 'binary_relation'],
         ['>', 'binary_relation'],
+        [$.keyword_like, 'pattern_matching'],
+        [seq($.keyword_not, $.keyword_like), 'pattern_matching'],
```
DerekStride, line 1344:
```suggestion
        [$._not_like, 'pattern_matching'],
```

grammar.js
```diff
@@ -1334,6 +1340,10 @@ module.exports = grammar({
...
+        [$._similar_to, 'pattern_matching'],
+        [seq($.keyword_not,$.keyword_similar, $.keyword_to), 'pattern_matching'],
```
DerekStride, line 1346:
```suggestion
        [$._not_similar_to, 'pattern_matching'],
```

grammar.js
```diff
@@ -19,6 +19,7 @@ module.exports = grammar({
       'binary_relation',
       'binary_concat',
       'clause_connective',
+      'pattern_matching',
```
DerekStride, line 22: These two need to switch right now they are grouping the predicates like this because `LIKE` is given higher precedence than `AND`.

```sql
WHERE a LIKE ('%a' AND a) NOT LIKE '%a'
```

```suggestion
      'pattern_matching',
      'clause_connective',
```

matthias-Q: Oh, ok. I was not aware that the order matters here.

DerekStride (approved):

PR #45, opened 2022-10-04
SPM support

mattmassicotte:
This adds Swift bindings and Swift Package Manager (SPM) support. Parser is unchanged.

I've been going through and trying to add Swift bindings for a bunch of parsers. Would be happy to answer questions!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ build
 /examples/npm
 package-lock.json
 /target/
+/.build/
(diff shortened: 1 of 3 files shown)


DerekStride:
I have a few questions:

1. Is there anything that would need to be specifically maintained with these bindings?
2. Can you add some links to other PRs you've opened in the PR description, especially any with context you think relevant.
3. Was this auto-generated (and how was it generated) or created manually?

PR #51, opened 2022-11-17
Wrap select expressions in a term node

DerekStride:
## What

The main goal of this PR is to wrap the the select expression items in a new `term` node. This allows us to properly attribute the `alias` to a specific term in the list of select expressions.
  - The `Aliased` test in `test/corpus/casting.txt` is a good example of how this helps.

## Addtional Improvements

There's a lot of additional improvements in this PR. I've tried to isolated them by commit.

The first few commits are independent style changes.
1. Add a `compile` entry in the makefile to generate the `.so` parser file.
2. Auto format all the tests with `tree-sitter test --update`

Then I introduce the improvements to the grammar.
1. The `term` node.
2. I made the final semi-colon at the end of a statement optional.


dmfay (approved):
my first thought was "wait, aren't there a couple other places we'd want to apply this?" but on looking it over I don't think there are any since relation aliases are a different animal. LGTM

PR #55, opened 2022-11-29
Derek/remove artifacts

### LINKED ISSUE #54: Update the repo to not require checking in the parser files
Right now everything generated via `tree-sitter generate` gets checked in, I'd like to change it so that only the grammar.js is checked into the repo and the artifacts can be generated by the `tree-sitter` tool but consumers of this parser.

DerekStride:
## What

closes #54

I'm removing the generated artifacts so they don't show up in git / diffs. They can still be generated via `tree-sitter generate` in this repo. If the artifacts are something that's desired we can setup a way to publish them instead via a Github Actions integration.


phsurette (approved):
Makes sense to me.

package.json
```diff
@@ -5,7 +5,8 @@
   "main": "bindings/node",
   "scripts": {
     "test": "tree-sitter test",
-    "install": "node-gyp rebuild"
+    "generate": "tree-sitter generate",
+    "compile": "node-gyp rebuild"
```
dmfay, line 9: `node-gyp rebuild` has a hard dependency on the `generate` step -- it blows up without the src/parser.c and bindings/node/binding.cc; it could `npm run generate && node-gyp rebuild` maybe?

we also do need to have the `install` lifecycle script step generate the artifacts and rebuild if they're no longer packaged for downstream dependencies

dmfay (changes requested):

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "bindings/node",
   "scripts": {
     "test": "tree-sitter test",
-    "install": "node-gyp rebuild"
+    "install": "tree-sitter generate && node-gyp rebuild"
```
DerekStride, line 8: > node-gyp rebuild has a hard dependency on the generate step -- it blows up without the src/parser.c and bindings/node/binding.cc; it could npm run generate && node-gyp rebuild maybe?
>
> we also do need to have the install lifecycle script step generate the artifacts and rebuild if they're no longer packaged for downstream dependencies

Does this address your concerns? It'll run the generate & rebuild during the install and I removed the explicit scripts.

dmfay: yes it should! Go ahead and merge this first, I've got to rebase arrays anyway.

clason:
You could also just add them to `.gitattributes` (see, e.g., https://github.com/vigoux/tree-sitter-viml/blob/master/.gitattributes)? This way users don't necessarily have to have the tree-sitter CLI installed.

PR #56, opened 2022-11-30
Unary predicates

DerekStride:
## What

Updates the `(where_expression)` node to only accept predicates and adds support for:

```
SELECT *
FROM my_table m
WHERE m.is_not_deleted;
```


phsurette (approved):
This is a nice improvement. 

For a minute I was thinking that the `left` and `right` operands of a predicate with a binary operator should also be predicates, they have the same attributes as the unary predicate you are introducing here. But the unary predicate really is a predicate, it implicitly asserts that the column is true and the `left` and `right` operands don't assert anything on their own, so this makes sense.

dmfay (approved):

Issue #52, opened 2022-11-22
Support `BEGIN ... END` blocks


ObserverOfTime:
`BEGIN ... END` blocks are used in [MySQL](https://dev.mysql.com/doc/refman/8.0/en/begin-end.html) and [MSSQL](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/begin-end-transact-sql?view=sql-server-ver16).

Here's a minimal example (which GitHub also fails to highlight properly):
```sql
BEGIN
    CREATE TABLE FOO (bar INT);
END;
```

```
keyword_begin [0, 0] - [0, 5]
ERROR [1, 4] - [2, 3]
  keyword_table [1, 11] - [1, 16]
```

DerekStride:
Is this something you'd be interested in tackling? PRs are always welcome.

ObserverOfTime:
Maybe I can try when I have some time to look into it.

DerekStride:
I'll close this for now, re-open it / open a new issue if you need more functionality added 👍

PR #57, opened 2022-11-30
Simple compound statement support

DerekStride:
## What

Address https://github.com/DerekStride/tree-sitter-sql/issues/52

This PR lays the ground work for supporting [Compound Statements](https://dev.mysql.com/doc/refman/8.0/en/begin-end.html) (see also [MSSQL](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/begin-end-transact-sql?view=sql-server-ver16)).

This doesn't include support for labels, while, loop, or repeat primitives.

@observeroftime


ObserverOfTime:
This works and covers my current use case, thank you!

PR #62, opened 2022-12-21
Add unary predicates that work with field predicates

DerekStride:
## What

As indicated in the tests, this PR fixes field predicates with unary operators e.g. NOT

```sql
SELECT *
FROM my_table m
WHERE NOT m.is_deleted
  AND NOT m.is_invisible;
```

## Considerations

At the moment we consider `NULL` to be a literal value but according to [this reference](https://forcedotcom.github.io/phoenix/index.html#expression)  `IS NULL` & `IS NOT NULL` are special cases of the `condition` term in the grammar (we use `predicate` in place of `condition` in our grammar but they are largely the same).

Screenshot 2022-12-21 at 2 38 18 PM

We have a couple strategies that are possible for the following where node:

```sql
WHERE NOT foo
  AND bar IS NOT NULL
```

1. Switch to the special case form from the grammar

```
(predicate
  left: (predicate
     operator: (keyword_not)
     operand: (field name: (identifier)) @foo
  operator: (keyword_and)
  right: (predicate
    operand: (field name: (identifier)) @bar
    operator: (is_not_null (keyword_is) (keyword_not) (keyword_null))))
```

2. Remove the `is_not` node. This is my preferred next steps.

```
(predicate
  left: (predicate
     operator: (keyword_not)
     operand: (field name: (identifier)) @foo
  operator: (keyword_and)
  right: (predicate
    left: (field name: (identifier)) @bar
    operator: (keyword_is)
    right: (predicate
      operator: (keyword_not) @unary_predicate
      operand: (literal (keyword_null)))))
```

3. Keep the grammar the same as it is now

```
(predicate
  left: (predicate
     operator: (keyword_not)
     operand: (field name: (identifier)) @foo
  operator: (keyword_and)
  right: (predicate
    left: (field name: (identifier)) @bar
    operator: (is_not (keyword_is) (keyword_not))
    right: (literal (keyword_null))))
```

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1381,6 +1381,7 @@ module.exports = grammar({
     _unary_predicate: $ => prec(0, choice(
       ...[
         [$.keyword_not, 'unary_not'],
+        [$.bang, 'unary_not'],
       ].map(([operator, precedence]) =>
         prec.left(precedence, seq(
           field('operator', operator),
@@ -1491,6 +1492,8 @@ module.exports = grammar({
       $._double_quote_string,
     ),
     _number: _ => /\d+/,
+
+    bang: _ => '!',
 
     identifier: $ => choice(
       $._identifier,
--- a/grammar.js
+++ b/grammar.js
@@ -10,6 +10,7 @@ module.exports = grammar({
 
   precedences: $ => [
     [
+      'is_not',
       'unary_not',
       'binary_exp',
       'binary_times',
@@ -152,7 +153,7 @@ module.exports = grammar({
 
     _similar_to: $ => seq($.keyword_similar, $.keyword_to),
     _not_similar_to: $ => seq($.keyword_not, $.keyword_similar, $.keyword_to),
-    is_not: $ => seq($.keyword_is, $.keyword_not),
+    is_not: $ => prec('is_not', seq($.keyword_is, $.keyword_not)),
     _not_like: $ => seq($.keyword_not, $.keyword_like),
     _temporary: $ => choice($.keyword_temp, $.keyword_temporary),
     _not_null: $ => seq($.keyword_not, $.keyword_null),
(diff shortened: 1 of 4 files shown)


dmfay:
Phoenix is an SQL layer over the top of HBase (BigTable family) so I have reservations about accepting its grammar as authoritative -- I don't know what all compromises they may have had to make. I dug up these other examples:

- [SQLite's railroad diagram for `expr`](https://www.sqlite.org/syntax/expr.html) is a little funny in that SQLite definitely supports `is [not] null` but the most obvious path shows `isnull`, `notnull`, and `not`-`null` tokens, suggesting that the parser actually follows the path immediately below and treats `null` as an expr.
- the Postgres docs don't have a railroad diagram here but the [grammar](https://github.com/postgres/postgres/blob/master/src/backend/parser/gram.y#L14586) uses dedicated tokens for `null` and other `is` tests.
- [Oracle's version](https://docs.oracle.com/cd/B14117_01/server.101/b10759/conditions006.htm#i1050977) is equivalent to Phoenix's.

Anyway! To my eyes, SQLite's taking your option 3 and the others are closer to option 1. Importantly, this means `select 1 is not 2` yields `1` (true) in SQLite and a syntax error in Postgres, out of RDBMSs I have immediately to hand. If we're after a general/permissive SQL grammar, as I am, imo that's an argument for sticking with status quo/option 3 over option 1.

None of the others imo points toward breaking `x is not (null|true|false)` into a nested double predicate, the binary `x is y` and the unary `not (null|true|false)`. Postgres' parser walks an operand down explicit `is`-`[not]`-`null`, `is`-`[not]`-`true`, etc paths, and I would guess that Oracle and Phoenix do as well but haven't checked. In other words, they've got not just an `is_not` rule but separate `is_not_null`, `is_not_true`, etc rules.

I wonder whether your option 2 could be a product of the unresolved redundancy between our `predicate` and `binary_expression`. Really a predicate is a binary (or unary!) expression that resolves to a boolean value, and you can e.g. select predicates just fine, hence why `predicate` is an option in our `_expression`. If we add not unary predicates but unary _expressions_ we could move all the relation & pattern-matching stuff out of `predicate` into `binary_expression`, so predicates are reduced to the connective-disjunctive tissue between unary or binary expressions. I think that'd make that area of the grammar a lot cleaner whether we go with a permissive `is [not] expr` or restrict the right-hand side of those expressions to certain standard values.

DerekStride:
Closed in favour of #63

PR #63, opened 2023-01-10
Remove predicate and merge into binary_expression

DerekStride:
## What

Alternative to #62

The goal of this grammar is to be permissive and be able to parse and make sense of many different SQL dialects. Given this goal we have don't have a clear canonical reference implementation to base our grammar on. There are many conflicting implementations and in the past I haven't compared them in-depth. Without a clear reference and conflicting implementations it leaves us free to choose the best structure for our needs.

> redundancy between our predicate and binary_expression. Really a predicate is a binary (or unary!) expression that resolves to a boolean value.

> If we add not unary predicates but unary expressions we could move all the relation & pattern-matching stuff out of `predicate` into `_expression`

Based on https://github.com/DerekStride/tree-sitter-sql/pull/62#issuecomment-1364591529 this PR implements these ideas. I've removed the `predicate` node from the grammar entirely instead relying on the `_expression` nodes. This is a breaking change but one I think we should consider.

## Context

When it comes to tree-sitter grammars the pattern of having binary & unary expressions as a choice rule under a anonymous _expression node seems to be the common behaviour. e.g.

```rust
/// Rust
-1 + 2;
// (binary_expression
//   left: (unary_expression (integer_literal))
//   right: (integer_literal))
```

```ruby
## Ruby
-1 + 2
# (binary
#   left: (unary operand: (integer))
#   right: (integer))
```

However, unlike Rust [^1] & Ruby [^2] which are imperative languages, SQL is a domain specific declarative language. The notion of a `predicate` node is present in many implementations[^3].

Although, some implementations, e.g. sqlite [^4] make liberal use expressions.

```sqlite3
$ sqlite3
SQLite version 3.39.4 2022-09-07 20:51:41
sqlite> SELECT 1 IS NOT NULL WHERE -8 + 8;
sqlite> SELECT 1 IS NOT NULL WHERE -8 + 7;
1
sqlite> SELECT 1 IS NULL WHERE -8 + 7;
0
```

> In the above example `-8 + 8` gets evaluated to `0` (false) and `-8 + 7` gets evaluated to `-1` (true, any non-zero number is truthy). Additionally, `1 IS NOT NULL` gets evaluated to `1` (true) and `1 IS NULL` to `0` (false).

Given all of the above I think we can justify representing the `predicate` nodes as regular `binary_expression` nodes in our grammar. It's permissive enough to support sqlite and it doesn't seem like the right place to inject the concept of a predicate into the grammar. The tools built on top of the grammar can encode the concept of a predicate if needed.

## Implementations in the Wild

See also https://github.com/DerekStride/tree-sitter-sql/pull/62#issuecomment-1364591529.

[^1]: [tree-sitter-rust grammar](https://github.com/tree-sitter/tree-sitter-rust/blob/f7fb205c424b0962de59b26b931fe484e1262b35/grammar.js#L964-L982)
[^2]: [tree-sitter-ruby](https://github.com/tree-sitter/tree-sitter-ruby/blob/7a1921bcfd90e3a04c1ad011059087aaf0168dd4/grammar.js#L896-L917)
[^3]: [ANTLR grammar based on MySQL reference](https://github.com/antlr/grammars-v4/blob/38857b99fbb6effa129e2d49182d3814013f8e46/sql/mysql/Positive-Technologies/MySqlParser.g4#L2571-L2590)
[^4]: [SQLite's railroad diagram for expr](https://www.sqlite.org/syntax/expr.html)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,26 +2,9 @@
 
 [![Build/test](https://github.com/derekstride/tree-sitter-sql/actions/workflows/ci.yml/badge.svg)](https://github.com/derekstride/tree-sitter-sql/actions/workflows/ci.yml)
 
-SQL grammar for [tree-sitter](https://github.com/tree-sitter/tree-sitter).
+A general/permissive SQL grammar for [tree-sitter](https://github.com/tree-sitter/tree-sitter).
 
 ## Features
-
-- [x] SELECT statements (some features may be incomplete)
-- [x] Function calls
-- [x] Index hints
-- [x] Comments
-- [x] Marginalia
-- [x] UPDATE statements
-- [x] DELETE statements
-- [x] CREATE statements
-- [x] INSERT/REPLACE statements
-- [x] DROP statements
-- [x] ALTER TABLE statements
-- [ ] CREATE INDEX statements
-- [ ] DROP INDEX statements
-- [ ] ALTER INDEX statements
-- [ ] EXPLAIN statements
-- [ ] Function definitions
 
 For a complete list of features see the the [tests](test/corpus)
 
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -2,35 +2,18 @@
...
-- [x] INSERT/REPLACE statements
-- [x] DROP statements
-- [x] ALTER TABLE statements
-- [ ] CREATE INDEX statements
-- [ ] DROP INDEX statements
-- [ ] ALTER INDEX statements
-- [ ] EXPLAIN statements
-- [ ] Function definitions
-
 For a complete list of features see the the [tests](test/corpus)
```
dmfay, line 9: I do like having the list but it'll take some time and care to produce a really useful one and keep it up to date, so this is fine for now imo

dmfay (approved):
this is a huge improvement! I'm about to push a branch based on this with some more changes (seeing as we're in a breaking-things mood) but everything in here lgtm

PR #64, opened 2023-01-11
more expression and other cleanup

dmfay:
Some more redundancies; feel free to exercise a line item veto here, I've itemized the commits.

I wanted to pull these out because they wrap _expressions at fixed or trivially findable positions in their respective clauses:

- join_expression: after `on`
- where_expression: always second
- order_expression: always third
- expression_list: always third (in partition and group)
- assignment_list: after `set`
- insert_expression: after `into`, basically split inserts in half

on column_list it just felt wrong that the most common use was anonymized and aliased while the somewhat obscure ordered syntax (not actually used in order_by, which has its own thing going on!) got pride of place. I've changed the alias to a more uniform `list`.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -762,10 +762,6 @@ module.exports = grammar({
     insert: $ => seq(
       choice($.keyword_insert, $.keyword_replace),
       $.keyword_into,
-      $.insert_expression,
-    ),
-
-    insert_expression: $ => seq(
       $.table_reference,
       optional(alias($._column_list_without_order, $.column_list)),
       choice(
--- a/grammar.js
+++ b/grammar.js
@@ -475,26 +475,12 @@ module.exports = grammar({
     ),
 
     expression_list :$ => seq(
-        $._expression_list,
+        $._expression,
         repeat(
             seq(
                 ',',
-                $._expression_list,
+                $._expression,
             )
-        )
-    ),
-
-    _expression_list: $ => seq(
-        choice(
-          $.literal,
-          $.field,
-          $.parameter,
-          $.case,
-          $.subquery,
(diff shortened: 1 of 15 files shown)


grammar.js
```diff
@@ -1316,15 +1264,13 @@ module.exports = grammar({
...
-        seq(
-          ',',
-          $.order_target,
+      seq(
+        $.order_target,
+        repeat(
+          seq(
+            ',',
+            $.order_target,
+          ),
```
DerekStride, line 1273:
```suggestion
      comma_list($.order_target, true)
```

DerekStride: Optionally, add a field e.g. "target" to make it a named node.

grammar.js
```diff
@@ -1296,15 +1246,13 @@ module.exports = grammar({
 
     where: $ => seq(
       $.keyword_where,
-      $.where_expression,
+      $._expression,
```
DerekStride, line 1249: Rather than have custom nodes for these, what do you think of using a named node, e.g. "expression" or "predicate"? Same for the former `join_expression', & 'order_expression', etc.

```suggestion
      field("expression", $._expression),
```

```suggestion
      field("predicate", $._expression),
```

DerekStride (approved):
I left a couple suggestions to improve a few things otherwise, all of these changes look good to me 👍

grammar.js
dmfay: I like it for join and where, definitely; order is kind of the odd one out because it's a list (potentially _containing_ binary expressions) while the others are binary expression trees linked by conjunctions and disjunctions. This one went places, though, so I'll raise that separately.

PR #65, opened 2023-01-12
Named expression node in join/where

dmfay:
This names `expression` fields in join-on and where `_expression`s so the root of a predicate tree has a consistent named field node. It does _not_ do this for expressions in the select, order, or assignment lists, or other places expressions can appear. That's something we might want to think about considering how `expression` is used in e.g. the SQLite grammar.

There was a fair bit of fallout from this, and some of it I don't entirely understand. Specifically, `table_reference`s following `from` suddenly wanted their nested `identifier`s to be `name:`d (cf `Parameterized Queries` tests); ditto `term`s in `select_expression`s (cf `Having with count function`). I think these may have been knock-on effects of the other big change I had to make here -- creating a no-op `operator` node as an alias target within `binary_expression`.

I did this because I noticed multi-keyword operations such as `is not distinct from` were, now that join and where expressions had become fields, resolving as multiple operators: `operator: (keyword_is) operator: (keyword_not) operator: (keyword_distinct)....`. Anonymous nodes wouldn't solve the problem since the tree generator passes directly through to their multiple children; so instead I aliased them all to `operator`s, saving only the connective and disjunctive keywords. I think it's more consistent overall.


test/corpus/expressions.txt
```diff
@@ -503,33 +513,36 @@ WHERE
...
                 right: (literal))
               operator: (keyword_and)
               right: (binary_expression
                 left: (field
                   name: (identifier))
-                operator: (keyword_not)
-                operator: (keyword_like)
+                operator: (operator
+                  (keyword_not)
+                  (keyword_like))
```
dmfay, line 530: this is another remaining weirdness: the `like` is resolved to an `(operator)` while `not like` is `(operator (keyword_not) (keyword_like))` (and you can see the `similar to` operations below behaving....similarly).

Replacing this line:
```
[$._not_like, 'pattern_matching'],
```
with
```
[seq($.keyword_not, $.keyword_like), 'pattern_matching'],
```
results in `not like` also resolving to an opaque `(operator)`. Should we try to prefer surfacing the keywords?

dmfay: I posit: yes, and it turned out to be pretty manageable at a second look -- even brought the total state count down by a few dozen in spite of introducing a few more node types!

DerekStride:
> the root of a predicate tree has a consistent named field node. It does not do this for expressions in the select, order, or assignment lists, or other places expressions can appear.

I think I'd prefer the field name of predicate over expression. It's more descriptive and indicative that the value will be coerced into a boolean.

> Specifically, table_references following from suddenly wanted their nested identifiers to be name:d (cf Parameterized Queries tests)

I think this is a behaviour from the `tree-sitter-cli` if the testcase doesn't include any field names then it's happy to ignore them. If the testcase specifies some field names then it seems to enter a "strict" mode.

>  did this because I noticed multi-keyword operations such as is not distinct from were, now that join and where expressions had become fields, resolving as multiple operators

This was actually an existing issue, see https://github.com/DerekStride/tree-sitter-sql/blob/3647b9f5b937269f43ac6e4b6ebcb6f52e033c17/test/corpus/expressions.txt#L525-L526

I think I'd prefer using dedicated named nodes instead of aliasing them all to an operator node with anonymous nodes. e.g. using `$.not_distinct_from` instead of `alias($._not_distinct_from, $.operator)`.

I've encapsulated the suggested changes in #66.

dmfay:
I did a little bit of precedence tweaking and brought `keyword_is` up to a named node -- if you're all good with that we can land it.

DerekStride (approved):

PR #66, opened 2023-01-12
Suggestions

DerekStride:
See https://github.com/DerekStride/tree-sitter-sql/pull/65 for discussion


dmfay (approved):
yeah I'm on board with all of this!

Issue #67, opened 2023-01-20
SQL highlighting issue in helix-editor, using tree-sitter-sql


LeoniePhiline:
The original bug report can be found at https://github.com/helix-editor/helix/issues/5186

Other than in the bug report, I am using now helix at latest master commit (edd0ba7 right now), but that makes no difference.

I updated `tree-sitter-sql` by overriding https://github.com/helix-editor/helix/blob/8e10c592b90383a90b7e4a9335d323f8fab99ab5/languages.toml#L1435-L1437 
followed by rebuilding grammers:

```
$ hx --grammar fetch                                                                                                                                                                                                                                                                                    
Fetching 128 grammars
127 up to date git grammars
1 updated grammars
        sql now on 6adc9578afb5c361858213adae5e2de990e46cb1


$ hx --grammar build                                                                                                                                                                                                                                                                                    
Building 128 grammars
127 grammars already built
1 grammars built now
        ["sql"]
```

however, the highlighting still fails, which makes me conclude this will rather be a tree-sitter-sql bug. (Or two, actually.)

## Summary

**_Example SQL:_**

`INSERT INTO ... SET ...` syntax:

```sql
INSERT INTO some_table
  field = "String does not get highlight in INSERT SET syntax";
```

helix edd0ba7 with tree-sitter-sql 9d9802991aa1d1bc00eee7713a838dab1eb4f149: String value does not get highlight:

![image](https://user-images.githubusercontent.com/22329650/208198460-1717581d-2cb4-4497-8a18-f2e81dffaddf.png)

`INSERT INTO ... (...) VALUES (...)` syntax with multiple values records:

```sql
INSERT INTO some_table
  (field)
VALUES
  ("String value"),
  ("String value");
```

helix edd0ba7 with tree-sitter-sql 9d9802991aa1d1bc00eee7713a838dab1eb4f149: Values of second record do not get highlight:

![image](https://user-images.githubusercontent.com/22329650/208145728-14550424-215e-4e2e-9ddf-2892e415cddf.png)

`GROUP_CONCAT` with `SEPARATOR`:

```sql
SELECT GROUP_CONCAT(uid SEPARATOR ",") 
FROM some_table 
GROUP BY some_field;

SELECT GROUP_CONCAT(uid SEPARATOR ",") 
FROM some_table 
GROUP BY some_field;
```

[...]

PR #68, opened 2023-01-20
Various improvements: aggregate functions, insert set syntax, & update join syntax

DerekStride:
## What

There are 3 updates that fix the issues described in https://github.com/DerekStride/tree-sitter-sql/issues/67.

cc// @LeoniePhiline

### Update insert to support SET syntax

https://github.com/DerekStride/tree-sitter-sql/pull/68/commits/3f08080faf4902ce2f528c4ae699fb42169131ae
https://github.com/DerekStride/tree-sitter-sql/pull/68/commits/2fdc42869aafb9d00d28ccaee6be8b10c0386f1b

Also updates the values branch of the grammar to allow for specifying multiple lists of items. The other commit includes support for the postgres style update statement.

### Rework special cases for aggregate functions

https://github.com/DerekStride/tree-sitter-sql/pull/68/commits/8c0ca86319fe745eebfe7d8ebdf6ac43100517ba
https://github.com/DerekStride/tree-sitter-sql/pull/68/commits/61c6e23337ed72464dc5c614fc47f876c1ad6ad6

Previous to this commit `count` was a special case node. It worked a lot like the `invocation` node but had some custom grammar rules.

This commit introduced another special case node, group_concat. According to the [MySQL reference](https://dev.mysql.com/doc/refman/8.0/en/aggregate-functions.html) there might not be many more like this but to provide a pattern for the future I moved both count & group_concat into a special anonymous node called `_aggregate_functions`.

The other commit adds the `_aggregate_expression` node from the suggestions.

### Consolidate UPDATE rules into 1 node

https://github.com/DerekStride/tree-sitter-sql/pull/68/commits/3b82416d4e4b10648cd79e2967b69e4f64480119

Previously there was separate anonymous nodes for single table & multi table update statements. This commit merges them into a single node and reworks them to support the JOIN syntax.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1016,21 +1016,22 @@ module.exports = grammar({
     count: $ => seq(
       field('name', alias($.keyword_count, $.identifier)),
       '(',
-      seq(
-        optional($.keyword_distinct),
-        field('parameter', choice($._expression, $.all_fields)),
-      ),
+      $._aggregate_expression,
       ')',
     ),
 
     group_concat: $ => seq(
       field('name', $.keyword_group_concat),
       '(',
-      optional($.keyword_distinct),
-      field('parameter', choice($._expression)),
-      optional($.order_by),
+      $._aggregate_expression,
       optional(seq($.keyword_separator, alias($._literal_string, $.literal))),
       ')',
+    ),
+
+    _aggregate_expression: $ => seq(
+      optional($.keyword_distinct),
+      field('parameter', choice($._expression, $.all_fields)),
+      optional($.order_by),
     ),
 
     invocation: $ => seq(
--- a/grammar.js
+++ b/grammar.js
@@ -802,11 +802,29 @@ module.exports = grammar({
     update: $ => seq(
       $.keyword_update,
       optional($.keyword_only),
-      comma_list($.relation, true),
-      repeat($.join),
-      $.keyword_set,
-      comma_list($.assignment, true),
-      optional($.where),
+      choice(
+        $._mysql_update_statement,
+        $._postgres_update_statement,
+      ),
+    ),
+
+    _mysql_update_statement: $ => prec(0,
+      seq(
+        comma_list($.relation, true),
+        repeat($.join),
+        $.keyword_set,
+        comma_list($.assignment, true),
+        optional($.where),
+      ),
+    ),
+
+    _postgres_update_statement: $ => prec(1,
+      seq(
+        $.relation,
+        $.keyword_set,
+        comma_list($.assignment, true),
+        optional($.from),
+      ),
     ),
 
     assignment: $ => seq(
--- a/grammar.js
+++ b/grammar.js
@@ -801,36 +801,12 @@ module.exports = grammar({
 
     update: $ => seq(
       $.keyword_update,
-      choice(
-        $._single_table_update,
-        $._multi_table_update,
-      ),
-    ),
-
-    _single_table_update: $ => seq(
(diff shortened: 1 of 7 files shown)


LeoniePhiline:
Wow @DerekStride, kudos to you!

This quick and comprehensive response by far exceeds all my expectations!

I will test your branch as well, to confirm the fixes.

grammar.js
```diff
@@ -787,38 +801,14 @@ module.exports = grammar({
...
-      optional($.where),
-      optional($.order_by),
-      optional($.limit),
-    ),
-
-    _multi_table_update: $ => seq(
-      $._table_references,
+      optional($.keyword_only),
+      comma_list($.relation, true),
+      repeat($.join),
```
dmfay, line 806: Postgres' update-join syntax is different, with the `set` clause coming before a more standard `from`-`join` construction that specifies the additional relations (the affected table is still specified right after the initial `update`). I think this is a net zero change for PG but I'll leave an example update statement with the tests if you want to try fitting that in too.

test/corpus/update.txt
```diff
@@ -125,11 +130,69 @@ SET ts = now();
...
+================================================================================
+Update with a JOIN
+================================================================================
+
+UPDATE table_a a
+INNER JOIN table_b b ON b.a = a.uid
+INNER JOIN table_c c ON c.b = b.uid
+SET
+a.d = 5;
+
```
dmfay, line 152: Postgres format:

```
UPDATE table_a as a
SET d = 5
FROM table_b b
INNER JOIN table_c c ON c.b = b.uid
WHERE b.a = a.uid;
```

grammar.js
```diff
@@ -1000,6 +990,11 @@ module.exports = grammar({
       ')',
     ),
 
+    _aggregate_function: $ => choice(
```
dmfay, line 993: great!

grammar.js
```diff
@@ -1010,6 +1005,16 @@ module.exports = grammar({
...
     ),
 
+    group_concat: $ => seq(
+      field('name', $.keyword_group_concat),
+      '(',
+      optional($.keyword_distinct),
+      field('parameter', choice($._expression)),
+      optional($.order_by),
+      optional(seq($.keyword_separator, alias($._literal_string, $.literal))),
+      ')',
```
dmfay, line 1015: in Postgres [this is possible with aggregate functions more generally](https://www.postgresql.org/docs/current/sql-expressions.html#SYNTAX-AGGREGATES)! e.g.

```
select count(distinct uid order by uid)
from table_a;
```

(although it doesn't really mean much with `count`)

I think putting this in `_aggregate_function` would be a good first step toward supporting that -- we can fill in the maxes mins etc and worry about extra stuff like the `filter` clause later.

dmfay (approved):
couple things to think about but it looks good in general!

grammar.js
DerekStride: done!

test/corpus/update.txt
DerekStride: https://github.com/DerekStride/tree-sitter-sql/commit/2fdc42869aafb9d00d28ccaee6be8b10c0386f1b

grammar.js
DerekStride: I've added an `_aggregate_expression` node to help generalize those cases: https://github.com/DerekStride/tree-sitter-sql/commit/61c6e23337ed72464dc5c614fc47f876c1ad6ad6

Issue #69, opened 2023-01-23
Support for Hive specific statements


matthias-Q:
Hi,
would it be possible to support Hive specific commands such `STORED AS PARQUET` or `COMPUTE STATS`?
This SQL extensions are supported by many databases these days (Redshift, Impala, Snowflake, Datafusion)

I would create a PR for that, if you are ok with that and once I have collected more examples.

Cheers,
Matt

matthias-Q:
It is a while since I have worked on this project. Do I still have to check in the parser.c and grammar.json/node-types.json? or is it sufficient to just check in the grammar.js?

DerekStride:
You should still check in the generated files for now we have a `.gitattributes` file to hide them from the GitHub diffs.

LeoniePhiline:
How do I trigger generating the files in `src/`?

matthias-Q:
@LeoniePhiline, I am using the rust based CLI for building the parser.c. You can install it via `cargo install tree-sitter-cli`

LeoniePhiline:
Thanks a lot!

LeoniePhiline:
@matthias-Q What do you think about the above build errors? They stem from use of `tree-sitter generate` CLI.

LeoniePhiline:
I got `v19.5.0`. 

But, after all, it seems to be irrelevant: `node-gyp rebuild` appears to be unnecessary to work on the grammar?! 🤔

PR #70, opened 2023-01-24
Parse more functionality of the `CREATE TABLE` statement

matthias-Q:
I noticed that currently, we cannot parse a statement like:

```sql
CREATE TABLE tab1 AS SELECT * FROM tab2
```
This is now possible. I also added the `ON COMMIT` behavior for temp tables and the `UNLOGGED` options.

Reference: https://www.postgresql.org/docs/15/sql-createtableas.html

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -524,8 +524,12 @@ module.exports = grammar({
       $.keyword_table,
       optional($._if_not_exists),
       $.table_reference,
-      $.column_definitions,
+      optional($.column_definitions),
       optional($.table_options),
+      optional(seq(
+          $.keyword_as,
+          $._select_statement,
+      )),
     ),
 
     create_view: $ => seq(
--- a/grammar.js
+++ b/grammar.js
@@ -527,15 +527,21 @@ module.exports = grammar({
       $.table_reference,
       choice(
           seq(
-              optional($.column_definitions),
+              $.column_definitions,
               optional($.table_options),
+              optional(
+                  seq(
+                      $.keyword_as,
+                      $._select_statement,
+                  ),
+              )
           ),
           seq(
               optional($.table_options),
-              optional(seq(
+              seq(
                   $.keyword_as,
                   $._select_statement,
-              )),
+              ),
           ),
       )
     ),
--- a/grammar.js
+++ b/grammar.js
@@ -865,14 +865,6 @@ module.exports = grammar({
         '=',
         field('value', $.identifier),
       ),
-      seq(
-          $.keyword_on,
-          $.keyword_commit,
-          choice(
-              $.keyword_drop,
-              seq($.keyword_delete, $.keyword_rows),
-              seq($.keyword_preserve, $.keyword_rows)
-          )),
     ),
 
     column_definitions: $ => seq(
--- a/grammar.js
+++ b/grammar.js
@@ -115,6 +115,7 @@ module.exports = grammar({
     keyword_owner: _ => make_keyword("owner"),
     keyword_temp: _ => make_keyword("temp"),
     keyword_temporary: _ => make_keyword("temporary"),
+    keyword_unlogged: _ => make_keyword("unlogged"),
     keyword_union: _ => make_keyword("union"),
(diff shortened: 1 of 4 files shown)


grammar.js
```diff
@@ -514,12 +516,21 @@ module.exports = grammar({
...
+      ),
       $.keyword_table,
       optional($._if_not_exists),
       $.table_reference,
-      $.column_definitions,
+      optional($.column_definitions),
       optional($.table_options),
+      optional(seq(
+          $.keyword_as,
+          $._select_statement,
```
dmfay, line 532: this should be a choice() between column_definitions and the `as select` clause rather than allowing both. The table_options I think follow column_definitions for MySQL if used so that'd be another seq, but doublecheck me on that one!

matthias-Q: Thanks. I have fixed that. I am planning to put a lot of the Hive specific statements (for #69) in table options.

I am not 100% sure about the MySQL reference, because the create_definition is wrapped in [] and ()...
I think I will test this out.

Reference: https://dev.mysql.com/doc/refman/8.0/en/create-table.html
```sql
CREATE [TEMPORARY] TABLE [IF NOT EXISTS] tbl_name
    [(create_definition,...)]
    [table_options]
    [partition_options]
    [IGNORE | REPLACE]
    [AS] query_expression
```

matthias-Q: This is a valid query in MySQL (but not in Postgres, even if MAX_ROWS is not there)

```sql
create table books2 
(a int)
MAX_ROWS=100
as select * from books
```

dmfay: oh interesting! I think we still need a `choice` because you _have_ to specify something, you can't just end it at `create table name` without specifying anything, but it'd be more like this:

```js
      ...,
      $.table_reference,
      choice(
          seq(
              $.column_definitions,
              optional($.table_options),
              optional(seq(
                $.keyword_as,
                $._select_statement,
              )),
          ),
          seq(
              optional($.table_options),
              seq(
                $.keyword_as,
                $._select_statement,
              ),
          ),
      )
```
so down the first branch you have to have the column definition and down the other you have to have `as select`.

matthias-Q: Oh, yeah. I had too many optional() in the last version.

One other thing, that is currently wrong is the `table_option` in line 855.
I have put the `ON COMMIT` in there. Should I create a new named node for that? There many more of these kinds of statements, that are optional, specified once after the "main" statement" and in a particular order. This gets really complicated, even when only considering only one dialect.

Also, `ON COMMIT` takes only effect when the table is `TEMP` or `TEMPORARY`.

dmfay: when to break out a named node is one of those more-art-than-science things but the two main reasons I ever have to do it are, one, it's reused elsewhere, and two, it represents a certain fuzzy level of internal complexity. The first doesn't apply here afaik, but I think you could make a good case for the second if you're inclined to do that.

Are you sure the options are ordered? I skimmed the 8.0 docs and they present table options alphabetically without an obvious mention of ordering.

When `on commit` does something is a distinct question from whether it parses -- as long as it's a no-op rather than a syntax error for permanent relations we're all good there. If it _is_ a syntax error in that case then I'm still not sure it's worth the effort of splitting temporary and permanent versions of `create table`.

matthias-Q: I have removed `on commit` for now, but kept the keywords. 
Regarding the other options mentioned above, I will include the partially when working on the hive specific commands for #69.

dmfay (approved):
looks good! Both this and @wishdev 's #75 need conflicts resolved and will cause conflicts for each other; if you read this first it'd be a good idea to comment on the other PR immediately so the author knows to wait.

wishdev:
I rebased #75 so it's good to go.

matthias-Q:
I fixed the conflicts.

LeoniePhiline:
Note: `(keyword_preserve)` is added but not used.

matthias-Q:
Yes, that is correct. I added it, but the correct implementation had some problems. Since it is not a very often used keyword, I implement not to use it for now.

Issue #71, opened 2023-01-26
Some more MariaDB syntax missing or inconsistent highlight


LeoniePhiline:
Using 4fe05b2d81565ddb689d2f415e07afdacc515c52

```sql
CREATE TABLE publications(
    `uid` INT AUTO_INCREMENT NOT NULL,
    `pid` INT DEFAULT 0 NOT NULL,
    `tstamp` INT UNSIGNED DEFAULT 0 NOT NULL,
    `crdate` INT UNSIGNED DEFAULT 0 NOT NULL,
    `cruser_id` INT UNSIGNED DEFAULT 0 NOT NULL,
    `deleted` SMALLINT UNSIGNED DEFAULT 0 NOT NULL,
    `hidden` SMALLINT UNSIGNED DEFAULT 0 NOT NULL,
    `notes` TEXT DEFAULT NULL,
    `code` VARCHAR(3) DEFAULT '' NOT NULL,
    `name` VARCHAR(100) DEFAULT '' NOT NULL,
    `frequency` SMALLINT UNSIGNED DEFAULT 0 NOT NULL,
    `editions` INT UNSIGNED DEFAULT 0 NOT NULL,
    `subscription_types` INT UNSIGNED DEFAULT 0 NOT NULL,
    `methods` INT UNSIGNED DEFAULT 0 NOT NULL,
    INDEX `parent`(pid),
    PRIMARY KEY(uid)
) DEFAULT CHARACTER SET utf8mb4 COLLATE `utf8mb4_unicode_ci` ENGINE = InnoDB;
```
![image](https://user-images.githubusercontent.com/22329650/214721214-e7923c5a-d095-4ab1-8c08-73a31574ed49.png)

1. `NOT` is sometimes lit, sometimes it isn't. ✅ _Fixed by  #73_
2. `UNSIGNED` might be something to highlight as keyword?  ✅ _Fixed by  #88_
3. `INDEX` should be a keyword like `KEY` ✅ _Fixed by  #77_
4. `DEFAULT CHARACTER SET`, `COLLATE` and `ENGINE` all keywords. ✅ _Fixed by  #77_

---

```sql
CREATE TABLE some_table(
    `code` VARCHAR(9) DEFAULT '' NOT NULL,
    `name` VARCHAR(100) DEFAULT '' NOT NULL,
    `published_at` DATE DEFAULT '0000-00-00' NOT NULL,
    INDEX `parent`(pid),
[...]

PR #73, opened 2023-01-29
Default values in Column Definitions

matthias-Q:
Fixes one problem reported in #71

Modified `column_definitions` to accept `DEFAULT` with `literal` or `invocation` optionally in parenthesis.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -868,7 +868,13 @@ module.exports = grammar({
     ),
     _inner_default_expression: $ => choice(
         $.literal,
+        $.list,
+        $.cast,
+        $.binary_expression,
+        $.unary_expression,
+        $.array,
         $.invocation,
+        alias($.implicit_cast, $.cast),
     ),
 
     column_definition: $ => seq(
--- a/grammar.js
+++ b/grammar.js
@@ -853,7 +853,22 @@ module.exports = grammar({
       ')',
     ),
 
+    column_definition: $ => seq(
+      field('name', $.identifier),
+      field('type', $._type),
+      repeat($._column_constraint),
+    ),
 
+    _column_constraint: $ => choice(
+        choice(
+            $.keyword_null,
+            $._not_null,
+        ),
+        $._default_expression,
+        $._primary_key,
+        $.keyword_auto_increment,
+        $.direction,
+    ),
 
     _default_expression: $ => seq(
         $.keyword_default,
@@ -875,19 +890,6 @@ module.exports = grammar({
         $.array,
         $.invocation,
         alias($.implicit_cast, $.cast),
-    ),
-
-    column_definition: $ => seq(
-      field('name', $.identifier),
-      field('type', $._type),
-      optional($._default_expression),
-      choice(
-        optional($.keyword_null),
-        optional($._not_null),
-      ),
-      optional($.keyword_auto_increment),
-      optional($._primary_key),
-      optional($.direction),
     ),
 
     constraints: $ => seq(
--- a/grammar.js
+++ b/grammar.js
@@ -853,13 +853,31 @@ module.exports = grammar({
       ')',
     ),
 
+
+
+    _default_expression: $ => seq(
+        $.keyword_default,
+            choice(
+            seq(
+                '(',
+                    $._inner_default_expression,
+                ')',
+            ),
+            $._inner_default_expression,
+        )
+    ),
+    _inner_default_expression: $ => choice(
+        $.literal,
+        $.invocation,
+    ),
+
     column_definition: $ => seq(
       field('name', $.identifier),
       field('type', $._type),
+      optional($._default_expression),
       choice(
         optional($.keyword_null),
         optional($._not_null),
-        optional($._default_null),
       ),
       optional($.keyword_auto_increment),
       optional($._primary_key),
--- a/src/grammar.json
+++ b/src/grammar.json
@@ -3891,7 +3891,36 @@
         },
         {
           "type": "SYMBOL",
+          "name": "list"
+        },
+        {
+          "type": "SYMBOL",
+          "name": "cast"
+        },
+        {
+          "type": "SYMBOL",
+          "name": "binary_expression"
+        },
+        {
+          "type": "SYMBOL",
+          "name": "unary_expression"
+        },
+        {
+          "type": "SYMBOL",
+          "name": "array"
+        },
+        {
+          "type": "SYMBOL",
           "name": "invocation"
+        },
+        {
+          "type": "ALIAS",
+          "content": {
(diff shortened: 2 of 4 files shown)


grammar.js
```diff
@@ -853,13 +853,31 @@ module.exports = grammar({
...
+    ),
+    _inner_default_expression: $ => choice(
+        $.literal,
+        $.invocation,
+    ),
+
     column_definition: $ => seq(
       field('name', $.identifier),
       field('type', $._type),
+      optional($._default_expression),
```
dmfay, line 877: these extra column constraints are order-independent (I can `create table aaaa(id int primary key not null default 1);`). I think this puts us in a somewhat difficult position. We could represent them as a `repeat` of zero or more constraints (the column constraint being a probably-anonymous node offering a choice between e.g. `_default_expression`, `choice(null, notnull)`, etc), thereby not tracking or restricting the quantity and allowing multiple defaults/nulls/notnulls.

Alternatively, we could encode the Cartesian product of the options. We'd have to check both `seq(default, choice(null, notnull))` and `seq(choice(null, notnull), default)`.

The [Postgres grammar for `create table`](https://www.postgresql.org/docs/current/sql-createtable.html) lumps column constraints into a single grammatical node

```
[ CONSTRAINT constraint_name ]
{ NOT NULL |
  NULL |
  CHECK ( expression ) [ NO INHERIT ] |
  DEFAULT default_expr |
  GENERATED ALWAYS AS ( generation_expr ) STORED |
  GENERATED { ALWAYS | BY DEFAULT } AS IDENTITY [ ( sequence_options ) ] |
  UNIQUE [ NULLS [ NOT ] DISTINCT ] index_parameters |
  PRIMARY KEY index_parameters |
  REFERENCES reftable [ ( refcolumn ) ] [ MATCH FULL | MATCH PARTIAL | MATCH SIMPLE ]
    [ ON DELETE referential_action ] [ ON UPDATE referential_action ] }
```

and will for example parse but refuse to execute a definition with `colname int default 1 default 2` or `colname int null not null`. I'm inclined to follow that, especially in light of the other option -- leave it to linters to analyze the syntax tree and check for conflicting options.

grammar.js
```diff
@@ -853,13 +853,31 @@ module.exports = grammar({
...
+                    $._inner_default_expression,
+                ')',
+            ),
+            $._inner_default_expression,
+        )
+    ),
+    _inner_default_expression: $ => choice(
+        $.literal,
+        $.invocation,
+    ),
```
dmfay, line 872: default expressions are a slightly bigger subset of `_expression` than this -- you can set `default 1+1`, cast with `default (5/2)::int`, or if you have a composite type representing an (int, text) tuple `default (1, 'tag')`. Many of the possibilities don't usually make a ton of sense but they _are_ legal :)

Out of `_expression` I think we need:

- $.literal
- $.list (for composite types)
- $.case (legal, kind of nonsensical)
- $.cast
- alias($.implicit_cast, $.cast)
- $.invocation
- $.binary_expression
- $.unary_expression
- $.array

The binary & unary expressions are kind of escape hatches into illegal behavior since they include all of `_expression` but I'm not sure there's any practical way around that.

matthias-Q: Ah SQL... *sigh*

Alright. I will just add all of these

Thanks for the research @dmfay

grammar.js
matthias-Q: I am not entirely sure I understood the second option. This is your preferred way of encoding it right?

matthias-Q: Something like this?

```
    _column_constraints: $ => choice(
        seq(
            choice($.keyword_null, $._not_null), 
            $._default_expression
        ),
        seq(
            $._default_expression, 
            choice($.keyword_null, $._not_null)
        )
    ),

    column_definition: $ => seq(
      field('name', $.identifier),
      field('type', $._type),
      optional($._column_constraints),
      // optional($._default_expression),
      // choice(
      //   optional($.keyword_null),
      //   optional($._not_null),
      // ),
      optional($.keyword_auto_increment),
      optional($._primary_key),
      optional($.direction),
    ),
```

2 issues:
1. this does not work at the moment, I get some ERRORs in the AST for some tests
2. Looking at the Postgres Documentation above this `_column_constrain` needs to have more things in it like e.g. `_primary_key`.

matthias-Q: I fiddled around a little bit and I think this is how is should be. I assume this was the 1st variant that you suggested.

However, this also does not work because of this error:
`The rule `_column_constraints` matches the empty string.`

We also have the `constraints` separated from the `_column_constraints`, which is imo probably fine.

```javascript
    column_definitions: $ => seq(
      '(',
      $.column_definition,
      repeat(
        seq(',', $.column_definition),
      ),
      optional($.constraints),
      ')',
    ),

    column_definition: $ => seq(
      field('name', $.identifier),
      field('type', $._type),
      optional($._column_constraints)
    ),

    _column_constraint: $ => choice(
        $._default_expression,
        $.keyword_null,
        $._not_null,
        $.keyword_auto_increment,
        $._primary_key,
        $.direction,
    ),
    _column_constraints: $ => repeat(
        $._column_constraint,
    ),

    _default_expression: $ => seq(
        $.keyword_default,
            choice(
            seq(
                '(',
                    $._inner_default_expression,
                ')',
            ),
            $._inner_default_expression,
        )
    ),
    _inner_default_expression: $ => choice(
        $.literal,
        $.list,
        $.cast,
        $.binary_expression,
        $.unary_expression,
        $.array,
        $.invocation,
        alias($.implicit_cast, $.cast),
    ),
```

dmfay: yes sorry, that could've been clearer, trying to represent all possible orders is not practical.

`repeat` is _zero or more_ which I think is what you're running into -- the best way to express what you're after is probably this:

```javascript
    column_definition: $ => seq(
      field('name', $.identifier),
      field('type', $._type),
      repeat($._column_constraint) // zero or more singular column constraints follow
    ),
```

you could also use `repeat1` in `_column_constraints` but as just shown there isn't a real need for it, and its behavior would be contingent on the `optional` in `column_definition` so spread across multiple sites.

matthias-Q: Yeah, I was a bit confused by optional and repeat. Now it works. Thanks.

But I am still interested how the alternative option would have looked like (the Cartesian product of the options)

dmfay: a big sprawling mess is what it'd have looked like -- to support e.g. four options we would need a `choice` between `seq`s (primary key, default, null/notnull, unique), (default, primary key, null/notnull, unique), (default, null/notnull, primary key, unique), (default, null/notnull, unique, primary key), and so on to cover every possible permutation.

dmfay (approved):

PR #77, opened 2023-02-01
MariaDB syntax

DerekStride:
## What

Address some issues defined in #71

* Fixes for index keyword in column definition constraints
* Add some more options for tables options to support `COLLATE`, `DEFAULT CHARACTER SET` & `keyword_engine`.

```sql
    INDEX `parent`(pid),
    INDEX `range`(publication, published_at)
) DEFAULT CHARACTER SET utf8mb4 COLLATE `utf8mb4_unicode_ci` ENGINE = InnoDB;
```

@LeoniePhiline


LeoniePhiline:
LGTM, thank you so much!

The reformatted `test/corpus/create.txt` makes review a little difficult (better in a separate PR?), but I think I caught all the actual changes.

LeoniePhiline:
Could you also make `$.keyword_column` optional in https://github.com/DerekStride/tree-sitter-sql/blob/main/grammar.js#L660 ?

See also https://mariadb.com/kb/en/alter-table/ (`ADD [COLUMN] ...`)

To make:
image
work like:
image

LeoniePhiline:
Looking great!

image

image

In the long term it would be great if all character sets, all engines etc. could be imported and recognized as builtin. But that would need to be automated to be feasible...

Issue #76, opened 2023-01-31
generated files and workflows


dmfay:
We [need to check in the generated parser files](https://github.com/DerekStride/tree-sitter-sql/pull/59) to support downstream use, but that gets us into merge conflicts every time someone cuts a new branch before someone else lands theirs. It's only happened a few times so far, but if the pace picks up we should consider an action that can build those on pull requests and squash merge.

We could commit after build+test on the branch in progress, but that seems like a bit much.

DerekStride:
I've had a few ideas for options on how to handle the build artifacts:

1. I was thinking we can start tagging commits with version numbers and generate GitHub releases that will publish the parser files as artifacts/assets. [Example release for a rubygem](https://github.com/Faveod/ruby-tree-sitter/releases/tag/v0.20.6.2).

2. On merge into the main branch we can build them and push them to a separate branch (github pages style).

Both of these options can be automated with a Github actions workflow.

matthias-Q:
This would be great. The merge conflicts on the parser.c are really annoying.

DerekStride:
closed by #100

PR #80, opened 2023-02-01
feat: Optional `COLUMN` in `ALTER TABLE` for `ADD`, `CHANGE` and `DROP`

### LINKED ISSUE #78: Make `COLUMN` optional in  `ALTER TABLE t ADD`
I will try to prepare my own PR for this.

Make `$.keyword_column` optional in https://github.com/DerekStride/tree-sitter-sql/blob/main/grammar.js#L660 

See also  (`ADD [COLUMN] ...`):
- https://mariadb.com/kb/en/alter-table/
- https://dev.mysql.com/doc/refman/8.0/en/alter-table.html
- https://www.postgresql.org/docs/current/sql-altertable.html

To make:
image
work like:
image

LeoniePhiline:
Fixes #78

# Before
image

# After
image

# Noob check

I changed `grammar.js`, then ran `tree-sitter generate`. `src/` was updated. 
Did I do that correctly? I haven't worked with tree sitter grammars yet.

I also ran `tree-sitter build-wasm` and `tree-sitter playground` to verify the fix.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -660,7 +660,9 @@ module.exports = grammar({
 
     add_column: $ => seq(
       $.keyword_add,
-      $.keyword_column,
+      optional(
+        $.keyword_column,
+      ),
       optional($._if_not_exists),
       $.column_definition,
     ),
@@ -668,7 +670,9 @@ module.exports = grammar({
     alter_column: $ => seq(
       // TODO constraint management
       $.keyword_alter,
-      $.keyword_column,
+      optional(
+        $.keyword_column,
+      ),
       field('name', $.identifier),
       choice(
         seq(
@@ -703,7 +707,9 @@ module.exports = grammar({
 
     drop_column: $ => seq(
       $.keyword_drop,
-      $.keyword_column,
+      optional(
+        $.keyword_column,
+      ),
       optional($._if_exists),
       field('name', $.identifier),
     ),
--- a/src/grammar.json
+++ b/src/grammar.json
@@ -2953,8 +2953,16 @@
           "name": "keyword_add"
         },
         {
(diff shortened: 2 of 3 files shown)


DerekStride:
>  changed grammar.js, then ran tree-sitter generate. src/ was updated.

Yeah this is correct, we have #76 open to help address that. It creates a lot of merge conflicts e.g. I merged #77 so you'll have to rebase this branch and re-generate the files.

The only thing missing is a test, you can add the example from your screenshot to this file:

https://github.com/DerekStride/tree-sitter-sql/blob/main/test/corpus/alter.txt

Here's the official `tree-sitter test` [documention](https://tree-sitter.github.io/tree-sitter/creating-parsers#command-test) if you have any other questions, post them here and I'll be happy to help.

Optionally you can run `tree-sitter test --update` to have the the tool generate the s-expression for the test. It'll also do the auto-formatting on the other tests so you might want to separate the changes into 2 commits like I did in #77 if you go that route. It's sometimes helpful to use `git add -p test` to stage formatting changes in chunks before committing your changes.

matthias-Q:
@LeoniePhiline not sure about your setup, but I run multiplexed in multiple terminals a filewatcher like `watchexec -e txt,js make test` and in another terminal  my editor (neovim).

So as soon as I save a test or the grammar.js a new parser.c is generated and the tests are run. I also have another `watchexec` watching on a playground.sql with runs `watchexec -e sql tree-sitter parse playground.sql` in order to checkout. I found the wasm/browser interface a bit sluggish.

LeoniePhiline:
@DerekStride Rebased, added tests, and ready to merge! 🚀

PR #86, opened 2023-02-01
Add `unsigned int|integer`

### LINKED ISSUE #79: `UNSIGNED` is missing
Causing queries like this to fail:

```sql
SELECT CAST("1" AS UNSIGNED INTEGER);
```
_Example source: https://mariadb.com/kb/en/cast/ ._

Tree:

image

`'unsigned'` does not appear in `grammar.js` at all and should probably be a keyword and part of the numeric type declarations.

matthias-Q:
Fixes #79


DerekStride (approved):

grammar.js
```diff
@@ -197,6 +197,13 @@ module.exports = grammar({
     keyword_numeric: _ => make_keyword("numeric"),
     keyword_real: _ => choice(make_keyword("real"),make_keyword("float4")),
     keyword_float: _ => make_keyword("float"),
+    unsigned_integer: _ => seq(
+        make_keyword("unsigned"),
+        choice(
+            make_keyword("integer"),
+            make_keyword("int")
+        )
```
DerekStride, line 205: We already have a `keyword_int` to represent the choice of possible values. Could we use that here instead? Also, it might be good to have `unsigned` be a named node (e.g. `$.keyword_unsigned: _ => make_keyword("unsigned")`) 

```suggestion
        make_keyword("unsigned"),
        $.keyword_int,
```

LeoniePhiline:
Note that `UNSIGNED` not only applies to `INT`:

> Most numeric types can be defined as [...] UNSIGNED

https://mariadb.com/kb/en/numeric-data-type-overview/#signed-unsigned-and-zerofill

matthias-Q:
I will remove the `INT`. Looks like the usage of `UNSIGNED` is quite different for each dialect.

@DerekStride This was my first attempt, but while compiling the parser.c, I got an error that `$` was not found. I came across this issue a few months ago as well. Let me try again later.

grammar.js
```diff
@@ -197,6 +197,13 @@ module.exports = grammar({
     keyword_numeric: _ => make_keyword("numeric"),
     keyword_real: _ => choice(make_keyword("real"),make_keyword("float4")),
     keyword_float: _ => make_keyword("float"),
+    unsigned_integer: _ => seq(
```
DerekStride, line 200: Would it be this type of issue that causes `$` to raise saying it's not found?

```suggestion
    unsigned_integer: $ => seq(
```

matthias-Q:
Yes, that was the problem. Thanks!

DerekStride:
I've built on this PR in https://github.com/DerekStride/tree-sitter-sql/pull/88 to support some of the other syntax rules highlighted in https://github.com/DerekStride/tree-sitter-sql/pull/86#issuecomment-1413893890

DerekStride:
closed by #88 (commits included in that PR)

PR #87, opened 2023-02-02
Add Hive statements

### LINKED ISSUE #69: Support for Hive specific statements

matthias-Q:
This PR adds hive specific syntax for `CREATE TABLE` statements following the Impala syntax.

Hive style databases have an even more diverse set of SQL statements. I have tried to integrate them into the current syntax as smooth as possible. I have created a new node called `table_settings` with should include most of the new statements. Some of them are also supported by MySQL and Postgres (e.g. `PARTITION|PARTITIONED BY`). I also modified a bit the `TABLEPROPERTIES` since the `TBLPROPERTIES` of are quite similar.

Here is the reference statement that I tried to implement:
```SQL
CREATE [EXTERNAL] TABLE [IF NOT EXISTS] [db_name.]table_name
  (col_name data_type
    [constraint_specification]
    [COMMENT 'col_comment']
    [, ...]
  )
  [PARTITIONED BY (col_name data_type, ...)]
  [SORT BY ([column [, column ...]])]
  [ROW FORMAT row_format]
  [STORED AS file_format]
  [LOCATION 'hdfs_path']
  [CACHED IN 'pool_name' [WITH REPLICATION = integer] | UNCACHED]
  [TBLPROPERTIES ('key1'='value1', 'key2'='value2', ...)]
```
Note: i have excluded `COMMENT` and `SERDE_PROPERTIES` for now
Note2: Snowflake, Impala, AWS Athena, Hive, Datafusion (maybe BigQuery) share alot of these statements in slightly different fashion. I only have access to Impala, Athena and Snowflake for testing purposes.

Fixes #69 

Sorry for this huge PR. Next time I will try to cut it down into smaller pieces.

References:
https://impala.apache.org/docs/build/html/topics/impala_create_table.html

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -160,6 +160,37 @@ module.exports = grammar({
     keyword_unsigned: _ => make_keyword("unsigned"),
     keyword_zerofill: _ => make_keyword("zerofill"),
 
+    // Hive Keywords
+    keyword_external: _ => make_keyword("external"),
+    keyword_stored: _ => make_keyword("stored"),
+    keyword_cached: _ => make_keyword("cached"),
+    keyword_uncached: _ => make_keyword("uncached"),
+    keyword_replication: _ => make_keyword("replication"),
+    keyword_tblproperties: _ => make_keyword("tblproperties"),
+    keyword_compute: _ => make_keyword("compute"),
+    keyword_stats: _ => make_keyword("stats"),
+    keyword_location: _ => make_keyword("location"),
+    keyword_partitioned: _ => make_keyword("partitioned"),
+    keyword_comment: _ => make_keyword("comment"),
+    keyword_sort: _ => make_keyword("sort"),
+    keyword_format: _ => make_keyword("format"),
+    keyword_delimited: _ => make_keyword("delimited"),
+    keyword_fields: _ => make_keyword("fields"),
+    keyword_terminated: _ => make_keyword("terminated"),
+    keyword_escaped: _ => make_keyword("escaped"),
+    keyword_lines: _ => make_keyword("lines"),
+
+    // Hive file formats
+    keyword_parquet: _ => make_keyword("parquet"),
+    keyword_rcfile: _ => make_keyword("rcfile"),
+    keyword_csv: _ => make_keyword("csv"),
+    keyword_textfile: _ => make_keyword("textfile"),
+    keyword_avro: _ => make_keyword("avro"),
+    keyword_sequencefile: _ => make_keyword("sequencefile"),
+    keyword_orc: _ => make_keyword("orc"),
+    keyword_avro: _ => make_keyword("avro"),
+    keyword_jsonfile: _ => make_keyword("jsonfile"),
+
     // Operators
     is_not: $ => prec.left(seq($.keyword_is, $.keyword_not)),
     not_like: $ => seq($.keyword_not, $.keyword_like),
@@ -535,12 +566,22 @@ module.exports = grammar({
       ),
     ),
 
+    _table_settings: $ => choice(
+        $.table_partition,
+        $.stored_as,
+        $.storage_location,
+        $.table_sort,
+        $.row_format,
+    ),
+
+
     create_table: $ => seq(
       $.keyword_create,
       optional(
           choice(
               $._temporary,
-              $.keyword_unlogged
+              $.keyword_unlogged,
+              $.keyword_external,
           )
       ),
       $.keyword_table,
@@ -549,6 +590,7 @@ module.exports = grammar({
       choice(
           seq(
               $.column_definitions,
+              repeat($._table_settings),
               optional($.table_options),
               optional(
                   seq(
@@ -558,13 +600,14 @@ module.exports = grammar({
               )
           ),
           seq(
+              repeat($._table_settings),
               optional($.table_options),
               seq(
                   $.keyword_as,
                   $._select_statement,
               ),
           ),
-      )
+       ),
     ),
 
     create_view: $ => seq(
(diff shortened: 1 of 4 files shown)


LeoniePhiline:
I think `blubb.sql` was added by mistake?

blubb.sql
```diff
@@ -0,0 +1,6 @@
+create table test (a int not null) as select a from test2;
+
+-- STORED AS PARQUET
+-- LOCATION 'path'
+-- CACHED IN 'pool' with replication=2
+
```
DerekStride, line 6: > > I think blubb.sql was added by mistake?

Yeah, let's remove this file

playground.sql
```diff
@@ -0,0 +1 @@
+create table aaa (a int default null not null);
```
DerekStride, line 1: This file should be removed too

test.sql
```diff
@@ -0,0 +1,15 @@
...
+    (VALUES 
+        (true, true, false), 
+        (false, true, true), 
+        (true, true, true)) 
+    AS t (a,b,c)
+)
+SELECT 
+    CASE WHEN a THEN 'A' END AS A,
+    CASE WHEN b is true THEN 'B' END AS B,
+    CASE WHEN c is true THEN 'C' END AS C
```
DerekStride, line 14: This can be deleted too

test/corpus/create.txt
```diff
@@ -1,142 +1,137 @@
-================================================================================
+==================
 Create table
-================================================================================
+==================
 
 CREATE TABLE my_schema.my_table (id BIGINT NOT NULL PRIMARY KEY);
 
---------------------------------------------------------------------------------
+---
```
DerekStride, line 7: Looks like this is undoing the formatting that was done in https://github.com/DerekStride/tree-sitter-sql/pull/77.

You can use `make format` or `tree-sitter test --update` to auto-format these tests. I think it requires a failing test otherwise it skips the formatting, I usually change a test slightly to get it to apply to the file.

test/corpus/create.txt
```diff
@@ -805,80 +706,17 @@ CREATE UNLOGGED TABLE tableName (
         (column_definition
           (identifier)
           (numeric
-            (keyword_numeric)))))))
-
-================================================================================
-Create table with index
-================================================================================
```
DerekStride, line 812: Looks like this test was removed by mistake.

matthias-Q:
Oh, sorry. Yeah I added some stuff that was not meant to be added. I will fix this right away. I have to do a rebase anyway now.

test/corpus/create.txt
matthias-Q: Yes, I fixed that. Sorry.

test/corpus/create.txt
matthias-Q: Thanks for the hint. It is a bit dangerous since in autofills the expected value for the test. But still, good to know

test.sql
matthias-Q: Done

playground.sql
matthias-Q: done

blubb.sql
matthias-Q: Sorry. It is deleted now.

grammar.js
```diff
@@ -860,21 +903,139 @@ module.exports = grammar({
...
+        seq(
+            $.keyword_tblproperties,
+            '(',
+            $.table_option,
+            repeat(seq(',', $.table_option)),
+            ')',
+        ),
+        seq(
+            repeat1($.table_option)
+        ),
```
DerekStride, line 1028:
```suggestion
        repeat1($.table_option)
```

grammar.js
```diff
@@ -860,21 +903,139 @@ module.exports = grammar({
...
+
+    table_sort: $ => seq(
+        $.keyword_sort,
+        $.keyword_by,
+        seq(
+            '(',
+                $.identifier,
+                repeat(seq(',', ($.identifier))),
+            ')',
+        )
```
DerekStride, line 970:
```suggestion
        '(',
        $.identifier,
        repeat(seq(',', ($.identifier))),
        ')',
```

grammar.js
```diff
@@ -860,21 +903,139 @@ module.exports = grammar({
...
+                        choice(
+                            $.keyword_uncached,
+                            seq(
+                                $.keyword_with,
+                                seq(
+                                    $.keyword_replication,
+                                    '=',
+                                    field('value', alias($._number, $.literal)),
+                                ),
+                            ),
```
DerekStride, line 925:
```suggestion
                            seq(
                                $.keyword_with,
                                $.keyword_replication,
                                '=',
                                field('value', alias($._number, $.literal)),
                            ),
```

grammar.js
```diff
@@ -531,21 +572,23 @@ module.exports = grammar({
...
           seq(
+              repeat($._table_settings),
               optional($.table_options),
               seq(
                   $.keyword_as,
                   $._select_statement,
-              ),
-          ),
+              )
+          )
```
DerekStride, line 591:
```suggestion
              ),
          ),
```

grammar.js
```diff
@@ -531,21 +572,23 @@ module.exports = grammar({
...
           seq(
               $.column_definitions,
+              repeat($._table_settings),
               optional($.table_options),
               optional(
                   seq(
                       $.keyword_as,
                       $._select_statement,
-                  ),
+                  )
```
DerekStride, line 581:
```suggestion
                  ),
```

grammar.js
```diff
@@ -860,21 +903,139 @@ module.exports = grammar({
...
+                $.keyword_by,
+            ),
+        ),
+        choice(
+            seq('(', $.identifier, ')'), // postgres
+            $.column_definitions // impala/hive
+        )
+    ),
+
+    hive_stored_as: $ => seq(
```
DerekStride, line 996: I wonder if we should avoid named nodes with specific dialect names.

e.g. rename this to `stored_as` unless we have a conflict.

matthias-Q: yeah, I was unsure about this as well. I will remove the dialect hint from the name.

matthias-Q: Alternatively, I could make these nodes anonymous like `_mysql_update_statement` and `_postgres_update_statement`

DerekStride (approved):
This looks good to me, just need to rebase on main again

matthias-Q:
@DerekStride rebased & squashed

PR #88, opened 2023-02-02
Add unsigned & zerofill options to numeric types

DerekStride:
## What

Follow-up to #86 (includes the original commits)

This version of the PR adds an `unsigned_type` function that will create a choice rule to prefix a node with `UNSIGNED` or suffix it with either `UNSIGNED` or `ZEROFILL`.

It should cover most cases highlighted in the MariaDB & MySQL docs see https://github.com/DerekStride/tree-sitter-sql/pull/86#issuecomment-1413893890

https://mariadb.com/kb/en/numeric-data-type-overview/#signed-unsigned-and-zerofill
https://dev.mysql.com/doc/refman/8.0/en/numeric-type-syntax.html


DerekStride:
@matthias-Q Does this look like an okay solution to you?

matthias-Q:
Looks good to me. I was wondering if we should add comments in the code where we use dialect specific implementations like `int1`

DerekStride:
> I was wondering if we should add comments in the code where we use dialect specific implementations like int1

I'm wondering how to handle this as well.

Anonymous nodes are another approach (nodes that start with `_`) e.g.

```javascript
unsigned: $ => choice($._mysql_unsigned, $._postgresql_unsigned),
_mysql_unsigned: $ => seq(...),
_postgresql_unsigned: $ => seq(...),
```

LeoniePhiline:
When naming such things, it could be useful to determine first what is SQL standard and what is a server-specific extension.

Also, how to name non-standard extensions which are not specific to one implementation, but shared by multiple (but not all)?

PR #90, opened 2023-02-03
CI autogenerate and commit parser.c

### LINKED ISSUE #76: generated files and workflows

matthias-Q:
This is a suggestion for resolve #76 

I tried to test it on my branch and I think it works. We also need to add the `parser.c` in the `.gitignore`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,8 @@ jobs:
       - run: |
           git config --global user.name "github-actions[bot]"
           git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
-          git add src/parser.c
+          # git add src/parser.c
+          touch testfile
+          git add testfile
           git commit -m "[skip ci] auto generate parser.c"
           git push
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: Build/test
 on:
   push:
     branches:
-      - "main"
+      - "ci_autogenerate_parser"
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
     branches:
@@ -24,7 +24,7 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     needs: test
-    if: github.ref == 'refs/heads/main'
+    # if: github.ref == 'refs/heads/main'
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,6 @@ jobs:
     needs: test
     if: github.ref == 'refs/heads/main'
     steps:
-      - name: Generate parser.c
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,10 +15,10 @@ jobs:
       matrix:
         os: [macos-latest, ubuntu-latest, windows-2019]
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
         with:
-          node-version: 14
+          node-version: 18
       - run: npm install
       - run: npm test
   deploy:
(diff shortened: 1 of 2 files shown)


matthias-Q:
Here are some references:
https://github.com/orgs/community/discussions/26615
https://stackoverflow.com/questions/58139406/only-run-job-on-specific-branch-with-github-actions
https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/

.github/workflows/ci.yml
```diff
@@ -21,3 +21,19 @@ jobs:
...
       - run: npm test
+  deploy:
+    runs-on: ubuntu-latest
+    needs: test
+    if: github.ref == 'refs/heads/main'
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 14
```
LeoniePhiline, line 32: Security support ends in 2 months and 3 weeks
(30 Apr 2023)

https://endoflife.date/nodejs

IIRC node 18 (latest LTS) worked for you in building node-gyp? That should be used (Support ends in 2 years and 2 months
(30 Apr 2025))

matthias-Q: So you suggest this?
```yaml
- uses: actions/checkout@v3
- uses: actions/setup-node@v3
  with:
    node-version: 18
```

I saw the deprecation warning, but it stated Node12 not Node14

LeoniePhiline: Yes. And I suggest changing the node version a few lines above as well.

.github/workflows/ci.yml
```diff
@@ -21,3 +21,19 @@ jobs:
...
+      - run: npm install
+      - run: |
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git add src/parser.c
```
matthias-Q, line 37: We should probably also include
```yaml
    git add src/grammar.json
    git add src/node-types.json
```

LeoniePhiline: Just `git add src`

LeoniePhiline, line 37:
```suggestion
          git add src
```

.github/workflows/ci.yml
```diff
@@ -21,3 +21,19 @@ jobs:
...
+          git commit -m "[skip ci] auto generate parser.c"
```
LeoniePhiline, line 38:
```suggestion
          git commit -m "[skip ci] auto generate parser"
```

.github/workflows/ci.yml
```diff
@@ -21,3 +21,19 @@ jobs:
           node-version: 14
```
LeoniePhiline:
```suggestion
          node-version: 18
```

dmfay:
Unfortunately I think there is an issue with regenerating the parser in a new commit on `main`: there will be a (brief) window during which the state of `main` is inconsistent and consumers could wind up with unexpected behavior.

I think the ideal place to regenerate the files is in a final commit on the branch _before_ it's merged into `main`; I've done [a little experimenting with a label-based process on my fork](https://github.com/dmfay/tree-sitter-sql/blob/main/.github/workflows/merge.yml) but haven't got it working yet -- feel free to pick it up and run with it if you think it's promising!

matthias-Q:
Ah, I see. You are right, for the 2-3 minutes while the CI actions are running, we have an inconsistent code base.

Can you explain how your approach should work in theory? If a PR is accepted, it should be labled with `merge` and then it triggers the actions in `merge.yaml`, correct?

dmfay:
> Ah, I see. You are right, for the 2-3 minutes while the CI actions are running, we have an inconsistent code base.
> 
> Can you explain how your approach should work in theory? If a PR is accepted, it should be labeled with `merge` and then it triggers the actions in `merge.yaml`, correct?

yes, that's the idea! Once labeled it should merge `main` into the branch (without committing), check out the `main` version of the parser files (eliminating any conflicts outside grammar.js which must be manually resolved), rebuild them to include the branch changes, then commit, push, and merge the pull request, and finally clean up the label.

matthias-Q:
I think that your solution is better. What is currently not working?

matthias-Q:
I was looking in you version and into https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-conditionals

Did you tried this line with out the curly braces `{{...}}`? 
```
    if: github.event.label.name == 'merge'
```

matthias-Q:
@dmfay @DerekStride I think I solved the issue. Building upon @dmfay work, I added a `merge.yml`, which triggers a "push force" when the `merge` label has been let to the PR.

Unfortunately, I uses a third party Github Action: https://github.com/marketplace/actions/github-push

I have tested this on a private repository. You also need to allow the Github actions bot to have write permissions on the repo.

DerekStride:
I took a different approach in #100, I like the label approach if we still want the artifacts on the `main` branch but I think I'd prefer to keep them separate.

DerekStride:
I'm going to close this since #100 was merged, we can re-open it if we decide that #100 was the wrong decision in the future.

PR #92, opened 2023-02-14
feat: Parse float

### LINKED ISSUE #81: Support for floating point literals
`_number` is defined as integer:
```
    _number: _ => /\d+/,
```
https://github.com/DerekStride/tree-sitter-sql/blob/286e10c5bc5d1703ee8f9afb351165a9a6182be1/grammar.js#L1488

This causes a parse error in queries containing floating point numbers:

```sql
SELECT 3.5;
```

For a fix, the question is if `_number` should add an optional `(?:\.\d+)?` or if a secondary number type should be added.

`$._number` is used in quite a few places. Some of these places should probably allow floating point numbers (e.g. ranges), while others should not (like params to `parametric_type()`.

### LINKED ISSUE #91: Missing floating numbers
The parser does not handle `1.1` `.1` or `1.`

Float should be parsed in statements like:

```sql
select
1.1 as f
where 1.1 = 1.1
```

or

```sql
select
a * 1.1
from b
```

I will submit a PR, but I am struggling to put the node at the right location.

matthias-Q:
Fixes #81

I have added the regex. Where should the node go?
When I add them to L1573:
```javascript
    _expression: $ => prec(1,
      choice(
        $.literal,
        $._decimal_number,
        $.field,
        $.parameter,
        $.list,
        $.case,
        $.window_function,
        $.subquery,
        $.cast,
        alias($.implicit_cast, $.cast),
        $._aggregate_function,
        $.invocation,
        $.binary_expression,
        $.unary_expression,
        $.array,
      )
    ),
```

Then a lot of tests fail, because instead of matching on `all_fields` it matches to `binary_expression`. This has probably to do with the `prec()` of this node.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1670,6 +1670,7 @@ module.exports = grammar({
       $._double_quote_string,
     ),
     _number: _ => /\d+/,
+    _decimal_number: _ => /\d*[.]?\d*/,
     bang: _ => '!',
 
     identifier: $ => choice(
--- a/grammar.js
+++ b/grammar.js
@@ -1672,9 +1672,9 @@ module.exports = grammar({
     ),
     _number: _ => /\d+/,
     _decimal_number: $ => choice(
-        seq(".", $._number),
-        seq($._number, ".", $._number),
-        seq($._number, "."),
+        seq(optional("-"), ".", $._number),
+        seq(optional("-"), $._number, ".", $._number),
+        seq(optional("-"), $._number, "."),
     ),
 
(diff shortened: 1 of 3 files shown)


matthias-Q:
I added it to the `$.literal` node, and switched from a regex to a tree sitter construct.
Supports signed and unsigned floats.

DerekStride (approved):
Literal seems like the right spot for it and using the treesitter constructs looks great.

PR #94, opened 2023-02-15
feat(pg): add interval parsing

matthias-Q:
Postgres interval parsing like:
`select '1 day'::interval`
or
`SELECT timestamp with time zone '2005-04-02 12:00:00-07' + interval '24 hours'`

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1233,8 +1233,7 @@ module.exports = grammar({
       $._type,
     ),
 
-
-    _interval_definitions: $ => repeat1(
+    interval_definitions: $ => repeat1(
          $._interval_definition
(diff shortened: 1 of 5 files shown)


grammar.js
```diff
@@ -1231,6 +1233,55 @@ module.exports = grammar({
       $._type,
     ),
 
+
+    _interval_definitions: $ => repeat1(
```
DerekStride, line 1237: I think declaring this as a named node e.g. `interval_definitions` would be nice to have it show up in the syntax tree.

matthias-Q:
I made the `interval_definitions` node public and added a test case.

PR #107, opened 2023-02-23
Add BIT data type for SQL Server

alexklapheke:
Equivalent to BOOLEAN in other DBMSs; see:
<https://learn.microsoft.com/en-us/sql/t-sql/data-types/bit-transact-sql>

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -223,6 +223,7 @@ module.exports = grammar({
     keyword_false: _ => make_keyword("false"),
 
     keyword_boolean: _ => make_keyword("boolean"),
+    keyword_bit: _ => make_keyword("bit"),
 
     keyword_smallserial: _ => choice(make_keyword("smallserial"),make_keyword("serial2")),
     keyword_serial: _ => choice(make_keyword("serial"),make_keyword("serial4")),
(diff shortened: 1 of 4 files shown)


matthias-Q:
Instead of adding `$.keyword_bit` directly to the types I suggest you create a node:
```javascript
    bit: $ => choice(
        $.keyword_bit,
        seq(
            $.keyword_bit,
            prec(0, parametric_type($, $.keyword_varying, ['precision'])),
        ),
        prec(1, parametric_type($, $.keyword_bit, ['precision'])),
    ),
```

Then, you also have to create the varying keyword (which implicitly already exists)
Lines 244 ans following like:
```javascript
    keyword_varying: _ =>make_keyword("varying"),

    keyword_char: _ => choice(make_keyword("char"), make_keyword("character")),
    keyword_varchar: $ => choice(
      make_keyword("varchar"),
      seq(
        make_keyword("character"),
        $.keyword_varying,
      )
    ),
```

An then some tests needs adjusting. This should make the bit datatype also available for postgres, mysql and mariadb.

DerekStride:
Closed by #110

PR #110, opened 2023-03-02
feat: add bit datatype

matthias-Q:
This PR extends #107 and includes also the variations for Postgres, MySQL and MariaDB. Thus, merging #107 is not necessary anymore, since there was not feedback in over a week.


matthias-Q:
Hmpf, I though it had merged from latest master... I will rebase

PR #111, opened 2023-03-02
WIP: Less permissive parenthesis

### LINKED ISSUE #106: Usage of parenthesis is wrong
In `$.list` we are using `paren_list()` which wraps parenthesis around a comma separated list. Therefore, statements like:
```sql
select 1 + (2 * 3, 2 * 3)
```
are parsed, even though they are not allows in SQL (I have just tested it against Postgres. I am sure it is wrong in any SQL dialect)

Here is the AST for this statement:
```
(program [0, 0] - [1, 0]
  (statement [0, 0] - [0, 25]
    (select [0, 0] - [0, 25]
      (keyword_select [0, 0] - [0, 6])
      (select_expression [0, 7] - [0, 25]
        (term [0, 7] - [0, 25]
          value: (binary_expression [0, 7] - [0, 25]
            left: (literal [0, 7] - [0, 8])
            right: (list [0, 11] - [0, 25]
              (binary_expression [0, 12] - [0, 17]
                left: (literal [0, 12] - [0, 13])
                right: (literal [0, 16] - [0, 17]))
              (binary_expression [0, 19] - [0, 24]
                left: (literal [0, 19] - [0, 20])
                right: (literal [0, 23] - [0, 24])))))))))
```

matthias-Q:
This is attempt to fix #106 

@DerekStride @dmfay I am unsure if we want to make the parser less permissive, since the main goal is syntax highlighting and not linting. 
However, I have create a function that just wraps parenthesis around the node and modified the `$.list` node to only accept specific nodes instead of `$._expression`. 

Not sure if I handled all cases in line 1787 and following. Since the testing suit only checks on the happy path (there is not "should fail"), I left out the tests.

(Maybe it is worth thinking about refactoring the usage of `$._expression` as well, since it is used everywhere and might not be fitting is some cases.)

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1698,23 +1698,26 @@ module.exports = grammar({
     ),
 
     _expression: $ => prec(1,
-      choice(
-        $.literal,
-        $.field,
-        $.parameter,
-        $.list,
-        $.case,
-        $.window_function,
-        $.subquery,
-        $.cast,
-        alias($.implicit_cast, $.cast),
-        $._aggregate_function,
-        $.invocation,
-        $.binary_expression,
-        $.unary_expression,
-        $.array,
-        $.interval,
-      )
+        prec.left(
+            choice(
+                $.literal,
+                $.field,
+                $.parameter,
+                $.list,
+                $.case,
+                $.window_function,
+                $.subquery,
+                $.cast,
+                alias($.implicit_cast, $.cast),
+                $._aggregate_function,
+                $.invocation,
+                $.binary_expression,
+                parens($.binary_expression),
+                $.unary_expression,
+                $.array,
+                $.interval,
+            )
+        ),
     ),
 
     binary_expression: $ => choice(
@@ -1781,7 +1784,21 @@ module.exports = grammar({
       ')',
     ),
 
-    list: $ => paren_list($._expression),
+    list: $ => prec(1,
+        paren_list(
+            choice(
+                $.literal,
+                $.field,
+                $.parameter,
+                $.cast,
+                $._aggregate_function,
+                $.invocation,
+                $.implicit_cast,
+                $.array,
+                $.interval,
+            )
+        ),
+    ),
 
     literal: $ => prec(2,
       choice(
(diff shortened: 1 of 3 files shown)


DerekStride:
> Maybe it is worth thinking about refactoring the usage of $._expression as well, since it is used everywhere and might not be fitting is some cases.

This might lead us to a better design. Currently nodes like `$.window_function` and `$.subquery` are available as an expression.

Limiting what a valid expression node is and explicitly adding a custom `choice($._expression, $.window_function)` in places where it's valid will hopefully lead to expressions being less of a catch-all node.

dmfay:
the problem is that SQL expressions [_really are_ that polymorphous](https://www.sqlite.org/syntax/expr.html); you can find almost every element of that `choice` in the SQLite railroad diagram there and more besides, like `collate` and `raise`. `list` is the eighth item down, and it is indeed recursive -- a list of `expr`s. The differences:

- we add `parameter`s for prepared statements
- we separate out window and aggregate functions, both to yield more specific node names and because sometimes there's extra stuff like Postgres' aggregate `filter where`
- we support non-SQLite expressions like arrays and intervals (great work with those btw!)

There are some places we're using `_expression` where it can't have its full effect, for example in `alter_column` with `set default`. In practice you're limited to literals and function calls there, but Postgres _parses_ `set default count(*)` -- then tells you aggregate functions aren't allowed. Similarly window functions aren't allowed in the `having` clause, and so on. But it isn't the parser that determines that.

The `select 1 + (2 * 3, 2 * 3)` example of #106, as well, parses. It dies because there isn't an operator defined for addition over (int, record) -- but you could define one! In fact, watch this drive:

```sql
create or replace function add_int_record(num int, rec record)
returns int
language plpgsql
as $$begin
return num + sum(value::int) from json_each_text(to_json(rec));
end;$$;

create operator + (
  leftarg = int,
  rightarg = record,
  function = add_int_record
);

select 1 + (2 * 3, 2 * 3);
```

Postgres returns 13 :)

matthias-Q:
Thank for this answer @dmfay. This is really interesting. I feel like, I learn more about SQL here than by using it.


I will close the PR and the issue, since is it not necessary.

DerekStride:
> This is really interesting. I feel like, I learn more about SQL here than by using it.

Same here, thanks for all that context @dmfay 🥇

PR #117, opened 2023-03-05
feat: add binary between expression

matthias-Q:
This PR adds `between` expressions for where clauses like

`select a from b where a between '2022-01-01' and '2022-02-01'`
or
`select a from b where a not between 1 and 2`

@DerekStride `not between` is currently not working. Do you have any clue why? It is structured exactly the same as `not like` but it still gives an error.


grammar.js
```diff
@@ -1779,6 +1781,18 @@ module.exports = grammar({
...
+        [$.keyword_between, 'binary_between'],
+        [$.not_between, 'binary_between'],
+      ].map(([operator, precendence]) =>
+        prec.left(precendence, seq(
+          field('left', $.identifier),
+          field('operator', operator),
+          field('low', $.literal),
+          $.keyword_and,
+          field('high', $.literal)
+        ))
```
DerekStride, line 1794: I'm not sure if this answers your question, I can look into it more later but from looking at this I don't think `between` & `not between` should be `binary_expression` nodes. They might deserve their own nodes.

matthias-Q: I have made it an extra node. However, as soon as the `not` token in in there it fails because it detects the where clause as a `binary_expression`

matthias-Q:
Looks like
```sql
select a from b where a not in (select 1)
```
has the same issue and is not working correctly.

matthias-Q:
@DerekStride did you had time to take a look? I am bit stuck here.

DerekStride:
I haven't had a chance yet but I'll take a look later today

DerekStride:
I gave it a try and found this solution, it's not great but I think it'll work for now: https://github.com/DerekStride/tree-sitter-sql/commit/e9cf694a5b92b6cc220c03f71154adb978ac886d

grammar.js
```diff
@@ -1793,6 +1796,21 @@ module.exports = grammar({
       ),
     ),
 
+    between_expression: $ => choice(
+      ...[
+            [$.keyword_between, 'between'],
+            [seq($.keyword_not, $.keyword_between), 'not_between'],
+        ].map(([operator, precedence]) =>
+                prec.left(precedence, seq(
+                field('left', $.identifier),
```
DerekStride, line 1805: Another option, is to use field here and then try to detangle the precedences, but I couldn't get that to work:

```suggestion
                field('left', $.field),
```

src/parser.c
DerekStride, line 1: With #100 we don't need these checked into main anymore so they can be removed from the PR

DerekStride (approved):
Just the 1 change to remove the parser artifacts and this is good to go 👍

PR #118, opened 2023-03-06
Add additional clauses/keywords to the insert statement

wishdev:


grammar.js
```diff
@@ -163,6 +163,12 @@ module.exports = grammar({
     keyword_preserve: _ => make_keyword("preserve"),
     keyword_unsigned: _ => make_keyword("unsigned"),
     keyword_zerofill: _ => make_keyword("zerofill"),
+    keyword_on_conflict: _ => make_keyword("on conflict"),
```
matthias-Q, line 166: we already have `keyword_on`. Would be better to use `seq($.keyword_on, ...)

grammar.js
```diff
@@ -163,6 +163,12 @@ module.exports = grammar({
...
+    keyword_do_nothing: _ => make_keyword("do nothing"),
+    keyword_do_update: _ => make_keyword("do update"),
```
matthias-Q, line 168: create a `keyword_do` and reuse it, `keyword_update` already exists as well. Like above, create a sequence out of these.

matthias-Q (commented):
Please split up the compound keywords like `ON CONFLICT`

grammar.js
```diff
@@ -948,18 +954,44 @@ module.exports = grammar({
     ),
 
     insert: $ => seq(
-      choice($.keyword_insert, $.keyword_replace),
-      $.keyword_into,
```
matthias-Q, line 952: why is `keyword_into` now optional? For which dialect is this change?.

grammar.js
```diff
@@ -948,18 +954,44 @@ module.exports = grammar({
...
       ),
       choice(
         $._insert_values,
-        $._insert_set,
+        $._set_values,
+      ),
+      optional(
+        seq(
+          $.keyword_on_conflict,
+          optional(
```
matthias-Q, line 984: This optional is not necessary. Otherwise it would be possible to end the statement with `on conflict`

grammar.js
wishdev: Done

grammar.js
wishdev: Done

grammar.js
wishdev: MySQL does not require it - is has always been optional.

grammar.js
wishdev: Fixed

grammar.js
matthias-Q: Ah yes. Before your PR the grammar required an `INTO` keyword, but it looks like it is optional in MySQL and MariaDB as well.

PR #124, opened 2023-03-21
Update installation instructions

### LINKED ISSUE #120: Please keep the generated parser files in the repository
The upcoming Emacs 29 can be optionally compiled with the tree-sitter library, and comes with a user command that assists users in building and installing the grammar libraries they need.  However, that command assumes the C/C++ parser and scanner source files are available in the corresponding repositories, so as to avoid requiring Emacs users to have too many development tools installed.

The recent removal of the generated parser files from the repository will cause that Emacs command to fail; users will be unable to build and install the tree-sitter grammar for SQL using your repository, unless they install quite a few more tools and build the library manually, something that is a significant obstacle for many Emacs users.

So please add those generated files back to the repository.  The Emacs user community will thank you.

Thanks in advance.

DerekStride:
## What

Closes #120

Adding installation instructions on how to compile the parser for use with `libtree-sitter`.

Emacs users have a more manual process to install parsers, this section lays out how to manually compile the parser files and also continues to call out that the parser files exist on the `gh-pages` branch.

I'm open to adding Emacs specific information but tried to keep the README more general. I'm also open for enabling Wiki's on this repo and linking to specific instructions for each editor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,12 +5,43 @@
 
 A general/permissive SQL grammar for [tree-sitter](https://github.com/tree-sitter/tree-sitter).
 
-## Notes
+## Installation
 
-We don't commit the generated parser files to the main branch. Instead, you can find them on the
-[gh-pages](https://github.com/DerekStride/tree-sitter-sql/tree/gh-pages) branch. They are also hosted on the [GitHub
-pages site](https://derek.stride.host/tree-sitter-sql/). We're open to feedback & encourage you to [open an
-issue](https://github.com/DerekStride/tree-sitter-sql/issues/new) to discuss any problems.
+**We don't commit the generated parser files to the `main` branch.** Instead, you can find them on the
+[gh-pages](https://github.com/DerekStride/tree-sitter-sql/tree/gh-pages) branch. We're open to feedback & encourage you
+to [open an issue](https://github.com/DerekStride/tree-sitter-sql/issues/new) to discuss any problems.
+
+They are also hosted on the [GitHub pages site](https://derek.stride.host/tree-sitter-sql/) and available for download
+here:
+[github://derekstride/tree-sitter-sql/gh-pages.tar.gz](https://github.com/DerekStride/tree-sitter-sql/archive/refs/heads/gh-pages.tar.gz).
+
+*Plugin maintainers ensure you specify to specify the `HEAD` (or a specific revision) of the `gh-pages` branch when integrating
+with this project.*
+
+### Step 1: Download the parser files
+
+**Using `git`**
+```bash
+git clone https://github.com/DerekStride/tree-sitter-sql.git
+cd tree-sitter-sql
+git checkout gh-pages
+```
+
+**Using `curl`**
+```bash
+curl -LO https://github.com/DerekStride/tree-sitter-sql/archive/refs/heads/gh-pages.tar.gz
+tar -xzf gh-pages.tar.gz
+cd tree-sitter-sql-gh-pages
+```
+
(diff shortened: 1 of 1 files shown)


Eli-Zaretskii:
Thanks for the README update.

However, please be aware that the `gh-pages` branch includes *only* the generated parser files. It does *not* include the grammar sources and all the auxiliary files, which are held on the `main` branch. So if someone wants to distribute the binaries of the grammar library, and include the sources with the distribution, they will need to collect the sources from the `main` branch and the generated parser files from the `gh-pages` branch.  The commands you show to fetch the parser only get the generated files, without the sources. People who make a point of distributing the full sources will thus need additional commands to get stuff from both branches.

DerekStride:
> So if someone wants to distribute the binaries of the grammar library, and include the sources with the distribution, they will need to collect the sources from the main branch and the generated parser files from the gh-pages branch.

That's a great point, I'm not aware of any use case for that at the moment so we can address that in the future if needed. That said there would be a couple options to solve the issue:

1. Merge the contents of both branches like you suggested.
2. Regenerate the parser artifacts with `tree-sitter-cli` from the `main` branch.

README.md
```diff
@@ -5,12 +5,40 @@
...
+```
+
+**Using `curl`**
+```bash
+curl -LO https://github.com/DerekStride/tree-sitter-sql/archive/refs/heads/gh-pages.tar.gz
+tar -xzf gh-pages.tar.gz
+cd tree-sitter-sql-gh-pages
+```
+
+**Note:** We don't commit the generated parser files to the main branch. Instead, you can find them on the
```
matthias-Q, line 26: Maybe change the order of `Notes` and `Installation` since the reasoning why we have the gh-pages branch is written in `Notes`

README.md
```diff
@@ -5,12 +5,40 @@
...
+[github://derekstride/tree-sitter-sql/gh-page.tar.gz](https://github.com/DerekStride/tree-sitter-sql/archive/refs/heads/gh-pages.tar.gz).
+
+### Step 2: Compile the Parser
+
+Tree-sitter parsers need to be compiled as a shared-object / dynamic-library, you can enable this by passing the
+`-shared` & `-fPIC` flags to your compiler.
+
+```bash
+cc -shared -fPIC -I./src src/parser.c -o sql.so
+```
```
matthias-Q, line 41: Maybe add a section/paragraph for plugin maintainers that the best way would be to use the HEAD of `gh-pages`

(I wasn't sure these days what to put in the lockfile.json of nvim-treesitter, but HEAD of gh-pages should be the correct rev, right?)

matthias-Q (approved):

Eli-Zaretskii:
> I'm not aware of any use case for that at the moment so we can address that in the future if needed

The future is already here:

                       https://sourceforge.net/projects/ezwinports/files/tree-sitter-lang-20230309-w32-bin.zip/download
                       https://sourceforge.net/projects/ezwinports/files/tree-sitter-lang-20230309-w32-src.zip/download

DerekStride:
> The future is already here

I'd advise to use my suggested option 2 from my earlier comment: https://github.com/DerekStride/tree-sitter-sql/pull/124#issuecomment-1478355316

Cloning the repo and running `npm install` should regenerate the parser artifacts and simplify packaging up the sources with compiled binaries.

Ideally, I'd like to generate the binaries via github actions and publish them as well in the future. I'd also like to reach a point where we can generate official releases with all the artifacts but I wouldn't consider the grammar ready for that yet.

PR #126, opened 2023-03-22
Add support for parenthesized expressions & EXISTS subquery

DerekStride:
## What

There's three changes being made in this PR, each are contained to individual commits.
1. Support for parenthesized expression. e.g. wrapping expressions in `(` & `)` to group them.
2. Support for the [`EXISTS`](https://dev.mysql.com/doc/refman/8.0/en/exists-and-not-exists-subqueries.html) expression.
    - Includes subquery support for `IN`
    - Excludes `NOT IN`, there was conflicts with precedence that I want to sort out separately.
3. Support for [`ANY`, `SOME`, & `ALL`](https://www.postgresql.org/docs/current/functions-subquery.html)

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -125,6 +125,8 @@ module.exports = grammar({
     keyword_unlogged: _ => make_keyword("unlogged"),
     keyword_union: _ => make_keyword("union"),
     keyword_all: _ => make_keyword("all"),
+    keyword_any: _ => make_keyword("any"),
+    keyword_some: _ => make_keyword("some"),
     keyword_except: _ => make_keyword("except"),
     keyword_intersect: _ => make_keyword("intersect"),
     keyword_returning: _ => make_keyword("returning"),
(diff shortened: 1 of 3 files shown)


DerekStride:
FYI I've enabled branch protection to require at least 1 approval & all the test suites passing before merging a PR

grammar.js
```diff
@@ -1482,6 +1482,11 @@ module.exports = grammar({
       paren_list(field('parameter', $._expression)),
     ),
 
+    exists: $ => seq(
```
matthias-Q, line 1485: Postgres seems to extend this syntax a bit https://www.postgresql.org/docs/current/functions-subquery.html

Does it make sense to add `not in`, `any`, `some`... as well?
`in` does work already, but only for lists and not for subqueries.

matthias-Q (commented):
Parenthesis change LGTM
`exists` is good, but adding more options (like `not in` etc) would be great

DerekStride:
I've update the PR (& the PR description) to include the suggested changes.

The one exception was `NOT IN` ended up causing precedence issues. I can open a separate issue to explore that in the future.

matthias-Q:
LGTM

matthias-Q (approved):

PR #139, opened 2023-04-25
allow count in window functions

matthias-Q:
* adds `count` as an option for the aggregation in window functions

@DerekStride a general question regarding the `count` node. For other functions we have the `invocation`. Why do we have this separation? 
In Impala the count in window function can also be empty (haven't checked other dialects) like:
`count() over (partition by col)`

Do you think we should rethink how we handle `invocations`. Maybe a named node for all possibile functions (`now`, `avg`, `sum`, `date_trunc`... <endless list>)

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1645,7 +1645,10 @@ module.exports = grammar({
     ),
 
     window_function: $ => seq(
-        $.invocation,
+        choice(
+          $.invocation,
+          $.count,
+        ),
         $.keyword_over,
         choice(
             $.identifier,
--- a/test/corpus/window_functions.txt
+++ b/test/corpus/window_functions.txt
@@ -686,3 +686,39 @@ FROM
       (relation
         (table_reference
           (identifier))))))
+
+================================================================================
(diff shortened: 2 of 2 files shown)


DerekStride:
> For other functions we have the invocation. Why do we have this separation?

I can't quite remember but I'm guessing that I implemented count before the invocation node existed.

> Do you think we should rethink how we handle invocations. Maybe a named node for all possible functions (now, avg, sum, date_trunc... )

I'm not sure on what is better here, we do that for expressions, where we have an anonymous `_expression` node which a bunch of concrete named child nodes. I was leaning towards removing the `count` node and normalizing on `invocation` but it might be better to follow the pattern of expressions, it seems to be working well.

DerekStride (approved):

matthias-Q:
The easier option would be to have just the `invocation` node. I don't see why `count` should be special. This way it also covers all of the dialect specific functions.
Let me create an issue for that refactor. I might pick it up when I am back mid may.

PR #141, opened 2023-04-30
feat: create function

dmfay:
not perfect but it's (Postgres-flavored) progress!

@matthias-Q I ran into a conflict between `create function x () as 'select 1'` and `create table y (id int) 'param'='value'`. Setting left precedence on `create table` worked, and I have a test case with a string-bodied function creating a table with options, although non-quoted since quoting _within_ a quote-delimited functino body isn't there yet. I did reorganize table settings under table options since they seem like roughly the same thing, and added Databricks' explicit `options` while trying to get to the bottom of the conflict.

Otherwise:

- lateral cross joins don't take an `on`, so that's fixed
- I did look into adding a custom scanner for tracking dollar quote opening/closing a while back but didn't have much success. I've [pushed that branch](https://github.com/DerekStride/tree-sitter-sql/tree/dmf/dollar-quoting) in case it's a helpful start

This also opens the enormous can of worms that is the question of where we draw the line on procedural control structures from pl/pgsql and friends! I've omitted them here since this is already plenty long.

### DIFF
--- a/test/corpus/functions.txt
+++ b/test/corpus/functions.txt
@@ -414,7 +414,7 @@ return 1;
    (identifier)
    (keyword_returns)
    (int (keyword_int))
-   (function_language)
+   (function_language (keyword_sql))
    (function_body
     (keyword_return)
     (literal)))))
@@ -449,7 +449,7 @@ return 1;
      (keyword_text)))
    (keyword_returns)
    (int (keyword_int))
-   (function_language)
+   (function_language (keyword_sql))
    (function_body
     (keyword_return)
     (literal)))))
@@ -459,13 +459,14 @@ Function details
 ================================================================================
 
 create or replace function public.fn()
- returns int
- language sql
- immutable
- parallel safe
- strict
- cost 100
- rows 1
+  returns int
+  language sql
+  immutable
+  parallel safe
+  leakproof
+  strict
+  cost 100
+  rows 1
 return 1;
 
 --------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


test/corpus/functions.txt
```diff
@@ -391,3 +391,499 @@ WHERE (
...
+    (keyword_end)
+    (dollar_quote)))))
+
+================================================================================
+Variable declarations
+================================================================================
+
+create or replace function public.fn()
+  returns int
+  language plpgsql
```
DerekStride, line 686: I've never worked with functions in SQL before, or plpgsql. A few naive questions:

Would this make sense as a separate grammar that imports this one? Similar to how the `typescript` grammar imports the `javascript` grammar.

I'm happy to merge this into this parser but it feels like a different thing so I'm curious on if it's valuable to separate them to help manage complexity.

dmfay: `create function` is in the SQL standard, but I believe the function _body_ was originally defined as a string, until the more recent introduction of `begin atomic` for SQL-language (that is, without procedural extensions) functions.

An SQL grammar does need to support `create function` and the SQL-standard `return` etc for completeness' sake, but that's a good callout on the subject of superset grammars! The one caveat is that reserving stuff like `if`/`elsif`/`else` and `raise` for tree-sitter-plpgsql or a generic tree-sitter-procedural-sql would mean gaps in coverage, especially for the syntax highlighting use case.

DerekStride: > The one caveat is that reserving stuff like if/elsif/else and raise for tree-sitter-plpgsql or a generic tree-sitter-procedural-sql would mean gaps in coverage, especially for the syntax highlighting use case.

We'd essentially need a to capture the `(function_body)` node use the separate parser to parse that node specifically. In neovim we'd use an `injection` query to do that.

For now, let's merge this in but if this part of the code grows complex we can investigate splitting it into a secondary parser.

DerekStride (approved):

PR #143, opened 2023-05-15
removed `count` in favour of `invocation`

matthias-Q:
This PR removed the `count` node in favor of the more generic `invocation.

I have a question:
We have `_expression`, `term` and `_select_expression` (which is `term` or `all_fields`). I may have forgotten this, but why do we have these? They look almost identical and offer only another layer in the AST hierarchy.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -82,7 +82,6 @@ module.exports = grammar({
     keyword_distinct: _ => make_keyword("distinct"),
     keyword_constraint: _ => make_keyword("constraint"),
     keyword_cast: _ => make_keyword("cast"),
-    keyword_count: _ => make_keyword("count"),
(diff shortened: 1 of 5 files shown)


DerekStride:
> We have _expression, term and _select_expression (which is term or all_fields). I may have forgotten this, but why do we have these? They look almost identical and offer only another layer in the AST hierarchy.

I think these have converged recently. We could remove `_select_expression` and just use an explicit `choice` rule or see if it's possible to add `term` to `_expression`.

matthias-Q:
Now that I think a little bit about it, I think `_select_expression is fine, since it combines `term` and `all_fields`, which is used multiple times.

`term` adds the optional alias and `_expression` is also fine, since it collects all the different variations (columns, subqueries etc)

PR #144, opened 2023-05-16
WIP: add multiple alter statements add once (t-sql)

### LINKED ISSUE #135: T-SQL Alter Table
TSQL:
```sql
ALTER TABLE Class_Session
ADD
type varchar(255) NOT NULL DEFAULT ('online');
```

Error:
```
Expected [A-Za-z0-9_:] or [A-Za-z0-9_] but " " found. (sql)
```

Expected:
No error, this is valid sql

image

matthias-Q:
Closes #135

This is an attempt to allow for multiple `alter table` `add` statements, as it is supported by T-SQL.

This does not work at the moment. I need a right precedence on the repeated section for at the `add_column` node, but this breaks the `_alter_specification` one hierarchy level up. I have currently not solution for this problem. @DerekStride, do you have an idea? The issue is that there can be multiple `_alter_specification` that are separated by commas, same as the `_inner_add_column` node that I have added.

Reference:
https://learn.microsoft.com/de-de/sql/t-sql/statements/alter-table-transact-sql?view=sql-server-ver16

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1032,8 +1032,8 @@ module.exports = grammar({
             seq(
               ",",
               $._alter_specifications
-            )
-          )
+            ),
+          ),
         ),
       ),
     ),
@@ -1051,14 +1051,26 @@ module.exports = grammar({
       $.change_ownership,
     ),
 
-    add_column: $ => seq(
-      $.keyword_add,
+    _inner_add_column: $ => seq(
       optional(
         $.keyword_column,
       ),
       optional($._if_not_exists),
       $.column_definition,
       optional($.column_position),
+    ),
+
+    add_column: $ => prec.right(
+      seq(
+        $.keyword_add,
+        $._inner_add_column,
+        repeat(
+          seq(
+            ',',
+            $._inner_add_column,
+          )
+        )
+      ),
     ),
 
     add_constraint: $ => seq(
--- a/test/corpus/alter.txt
+++ b/test/corpus/alter.txt
@@ -599,3 +599,44 @@ FOREIGN KEY ("accountId") REFERENCES "Account" ("accountId") ON DELETE CASCADE;
           (keyword_on)
           (keyword_delete)
           (keyword_cascade))))))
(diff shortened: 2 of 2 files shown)


matthias-Q:
Thanks, @dmfay for the suggestion. I will try this later today

matthias-Q:
@dmfay I have issues fixing the test alter.txt L428
```
ALTER TABLE IF EXISTS my_table
  ADD COLUMN IF NOT EXISTS val4 DATE,
  ALTER COLUMN val5 DROP NOT NULL, -- comment, ignore me!
  RENAME COLUMN val6 TO val7,
  DROP COLUMN IF EXISTS val8;
```

I have checked this statement in Postgres and Mysql and in both dialects it is not valid. The `rename` is the issue. Is this a bug in our implementation, or is is valid in another dialect? 

This is very confusing. When just running the alter table with only the renaming part, if works fine.

dmfay:
> @dmfay I have issues fixing the test alter.txt L428
>
> ```
> ALTER TABLE IF EXISTS my_table
>   ADD COLUMN IF NOT EXISTS val4 DATE,
>   ALTER COLUMN val5 DROP NOT NULL, -- comment, ignore me!
>   RENAME COLUMN val6 TO val7,
>   DROP COLUMN IF EXISTS val8;
> ```
>
> I have checked this statement in Postgres and Mysql and in both dialects it is not valid. The `rename` is the issue. Is this a bug in our implementation, or is is valid in another dialect?
>
> This is very confusing. When just running the alter table with only the renaming part, if works fine.

It looks like that was added [here](https://github.com/DerekStride/tree-sitter-sql/commit/be20ed053a6d0a6125e485f4796a488f91df987d) along with enabling more combinations of actions under `alter table` for Postgres, but you're right, Postgres' `alter table` allows column renames only as a single action. I don't know if any other dialects do allow it as one of multiple actions.

matthias-Q:
Closed by #175

Issue #145, opened 2023-05-17
select statements enclosed in parentheses next to 'union' aren't parsed proplerly


Appalled:
Hi, the following query isn't parsed correctly, please have a look.
`
-- CREATE TABLE tb AS
(
  SELECT 1 as col
)
UNION ALL
(
  SELECT 2 as col
)
;

Appalled:
Thanks. Maybe I misled you. It works after a create table statement, but fails alone. The following query is also legal.
"--" stands for comments in sparksql
```
(
  SELECT 1 as col
)
UNION ALL
(
  SELECT 2 as col
);
```

matthias-Q:
Ah, ok. Yeah I was wondering about that. I will try to fix this later today. I was on the notion that the select with parens should already work. However, the create table statement was also not working, so now it does :-)

`--` is the general comment string in all SQL dialects.

matthias-Q:
OK, I did take a deeper look into that, and I have to say that it is probably quite difficult to add to the parser in its current state. Adding these parenthesis for `select` statements creates a lot of other problems and conflicts. They can also around CTEs. But we have put CTEs in the node `_dml_statement`, which also includes `insert`, `delete` and `update`. Adding the


So...
```sql
(
  with x as (
      select 1 as a
  )
  select * from x
)
```
is valid, also a valid query.

I am wondering what the parenthesis should do anyway? They do not seem to have any purpose here and I can not think of a use case.


@DerekStride @dmfay what is your take on this?

matthias-Q:
I think I found a solution, but it involved adding a lot of precedence and associativity rules. I will create a PR later.

PR #149, opened 2023-05-24
feat: add parenthesized select and cte select statements

matthias-Q:
Addresses a remaining issue in #145 

It works, but I am not so super happy with it. I had to enter alot of associativity rules. Maybe that was necessary, since the change is so high up in the AST and it affects parenthesis.

Also, I am always about creating the anonymous nodes to reduce code replication like in `_cte_select_variants`.

P.S.
For MySQL, this is documented: https://dev.mysql.com/doc/refman/8.0/en/parenthesized-query-expressions.html

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -409,11 +409,11 @@ module.exports = grammar({
       $.keyword_regtype,
     ),
 
-    tinyint: $ => unsigned_type($, parametric_type($, $.keyword_tinyint)),
-    smallint: $ => unsigned_type($, parametric_type($, $.keyword_smallint)),
-    mediumint: $ => unsigned_type($, parametric_type($, $.keyword_mediumint)),
-    int: $ => unsigned_type($, parametric_type($, $.keyword_int)),
-    bigint: $ => unsigned_type($, parametric_type($, $.keyword_bigint)),
+    tinyint: $ => prec.left(unsigned_type($, parametric_type($, $.keyword_tinyint))),
+    smallint: $ => prec.left(unsigned_type($, parametric_type($, $.keyword_smallint))),
+    mediumint: $ => prec.left(unsigned_type($, parametric_type($, $.keyword_mediumint))),
+    int: $ => prec.left(unsigned_type($, parametric_type($, $.keyword_int))),
+    bigint: $ => prec.left(unsigned_type($, parametric_type($, $.keyword_bigint))),
 
     bit: $ => choice(
         $.keyword_bit,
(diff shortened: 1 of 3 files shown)


dmfay:
I think it might be a lot cleaner to make the parenthesized statement a choice further up. The examples `(select 1) limit 1` in the mysql docs mean `program` is probably out, but `select` or `select_expression` might work. Something like this --

```javascript
    _select_statement: $ => seq(
      choice(
        seq(
          $.select,
          optional($.from),
        ),
        seq(
          '(',
          $.select,
          optional($.from),
          ')'
        )
      ),
      ....
```

matthias-Q:
Not sure that will work, since also CTEs can be wrapped in parenthesis and `cte` and `_select_statement` are part of `_dml_statement`.

(One question: by "up in the tree" you mean further away from the root, right? thats the CS definition of a tree - in my head is the opposite :-D. But now that I write this it sounds more local.)

dmfay:
> Not sure that will work, since also CTEs can be wrapped in parenthesis and `cte` and `_select_statement` are part of `_dml_statement`.

yeah, that's a problem. Wrapping in parentheses is an error for insert/update/delete (in Postgres) so a `parenthesized_statement` would include only select with optional ctes. That might fit into `_dml_statement` though?

> (One question: by "up in the tree" you mean further away from the root, right? thats the CS definition of a tree - in my head is the opposite :-D. But now that I write this it sounds more local.)

oh I meant closer to the root! always tough with upside-down trees

matthias-Q:
Ok, but this is what I did. I added the variants with parenthesis to an anoym node in `_dml_statement` and not (further away from the root) in `_select_statement`

dmfay:
serves me right for multitasking! I think this might go somewhere:

```
    statement: $ => seq(
      choice(
        $._ddl_statement,
        $._dml_read,
        $._dml_write,
        seq(
          '(',
          $._dml_read,
          ')',
        ),
      ),
      optional(';'),
    ),
```

but right now I'm seeing more conflicts:

```
Unresolved conflict for symbol sequence:

  select  keyword_from  identifier  •  '('  …

Possible interpretations:

  1:  select  keyword_from  (invocation  identifier  •  '('  ')')
  2:  select  keyword_from  (invocation  identifier  •  '('  _expression  ')')
  3:  select  keyword_from  (invocation  identifier  •  '('  _expression  invocation_repeat1  ')')
  4:  select  keyword_from  (table_reference  identifier)  •  '('  …

Possible resolutions:

  1:  Specify a higher precedence in `invocation` than in the other rules.
  2:  Specify a higher precedence in `table_reference` than in the other rules.
  3:  Specify a left or right associativity in `table_reference`
  4:  Add a conflict for these rules: `table_reference`, `invocation`
```

A while back I was trying to replace `table_reference` and functions with a single `object_reference` (since we don't currently support `invocation`s that specify a schema, `select schemaname.fnname()`); maybe it's time to dust that off.

DerekStride (dismissed):
LGTM, all the parametric types have parenthesis so the associativity rules probably help the parser understand what's allowed where.

DerekStride:
Missed the conversation while reviewing the code first 🤦

PR #150, opened 2023-05-25
refactor: consolidate table_reference and invocation identifier into object_reference

dmfay:
Incorporates #149 ; not ready for primetime yet, there are four test failures that mostly stem from incorrect precedence negotiation between `object_reference` and `field` -- give it a whirl if you have time!

@matthias-Q if this is too much (it's a lot) you can avoid a lot of the precedence-setting in #149 by setting it in `parametric_type` as here. It still wants the precedence on `relation` but I haven't gone too far down that rabbit hole yet.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -9,7 +9,7 @@ module.exports = grammar({
   ],
 
   conflicts: $ => [
-    [$.object_reference, $.field],
+    [$.object_reference, $._qualified_field],
     [$.object_reference],
   ],
 
@@ -630,6 +630,7 @@ module.exports = grammar({
       field(
         'value',
         choice(
+          $._qualified_field,
           seq(
             optional(
               seq(
@@ -637,10 +638,7 @@ module.exports = grammar({
                 '.',
               ),
             ),
-            choice(
-              $.all_fields,
-              $.field,
-            ),
+            $.all_fields,
           ),
           $._expression,
         ),
@@ -1695,6 +1693,16 @@ module.exports = grammar({
 
     field: $ => field('name', $.identifier),
 
+    _qualified_field: $ => seq(
+      optional(
+        seq(
+          $.object_reference,
+          '.',
+        ),
+      ),
+      field('name', $.identifier),
+    ),
+
     implicit_cast: $ => seq(
       $._expression,
       '::',
@@ -2127,13 +2135,8 @@ module.exports = grammar({
     _expression: $ => prec(1,
       choice(
         $.literal,
-        seq(
-          optional(
-            seq(
-              $.object_reference,
-              '.',
-            ),
-          ),
+        alias(
+          $._qualified_field,
           $.field,
         ),
         $.parameter,
--- a/grammar.js
+++ b/grammar.js
@@ -630,16 +630,7 @@ module.exports = grammar({
       field(
         'value',
         choice(
-          $._qualified_field,
-          seq(
-            optional(
-              seq(
-                $.object_reference,
-                '.',
-              ),
-            ),
-            $.all_fields,
-          ),
+          $.all_fields,
           $._expression,
         ),
       ),
@@ -1644,7 +1635,15 @@ module.exports = grammar({
       optional($.direction),
     ),
 
-    all_fields: _ => '*',
+    all_fields: $ => seq(
+      optional(
+        seq(
+          $.object_reference,
+          '.',
+        ),
+      ),
+      '*',
+    ),
 
     parameter: $ => choice(
       "?",
--- a/grammar.js
+++ b/grammar.js
@@ -409,11 +409,11 @@ module.exports = grammar({
       $.keyword_regtype,
     ),
 
-    tinyint: $ => unsigned_type($, parametric_type($, $.keyword_tinyint)),
-    smallint: $ => unsigned_type($, parametric_type($, $.keyword_smallint)),
-    mediumint: $ => unsigned_type($, parametric_type($, $.keyword_mediumint)),
-    int: $ => unsigned_type($, parametric_type($, $.keyword_int)),
-    bigint: $ => unsigned_type($, parametric_type($, $.keyword_bigint)),
+    tinyint: $ => prec.left(unsigned_type($, parametric_type($, $.keyword_tinyint))),
+    smallint: $ => prec.left(unsigned_type($, parametric_type($, $.keyword_smallint))),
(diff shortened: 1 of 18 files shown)


grammar.js
```diff
@@ -625,7 +642,7 @@ module.exports = grammar({
       optional(
         $.keyword_only,
       ),
-      $.table_reference,
+      $.object_reference,
```
matthias-Q, line 645: We should not forget to fix the queries downstream in nvim-treesitter. There is a query matching on `table_reference`

matthias-Q (commented):
Looks good. I like the new `_dml_*` structure

matthias-Q:
The test `Update with a JOIN` is failing, because the `identifier` is missing the schema. In the test the assignment is written as `a.d = 5;`

matthias-Q:
I have a variant of your branch, where only one test fails (Simple select with schema and fully-pathed fields). Should I push it to this PR?

dmfay:
> I have a variant of your branch, where only two tests fail (Simple select with schema and fully-pathed fields, insert with field name). Should I push it to this PR?

yes absolutely!

> I do not understand why this [update assignment left-hand side] has been changed from field to identifier

I didn't follow up on MySQL multi-table updates (where the lhs can be qualified) until later and forgot to undo that change. I'll wait to change anything until you've pushed your updates though.

matthias-Q:
Not so many changes. I played around with your changes a bit.

The last two tests are breaking because of L1308 - `_set_values` is commented out. When adding this is a whole lot more breaks. I assume this is what you are referring to with the MySQL updates.

The other one is the fully qualified object_ref. LL1260 `object_reference` only has schema.name and it should be schema.table.col. But there I get a lot of issues.

matthias-Q:
Now I stumbled over your comments before `_set_value`. There is this test (Insert with field name) and it is valid in MySQL. I was not able to find it in the documentation but I did a quick test against a mysql instance.

dmfay:
some experimenting here in a toy subset https://github.com/dmfay/tsqltest

this gets sch.tbl.field, sch.tbl.* (missing a test in main atm!), etc correctly in select clauses but breaks join predicates, order columns, and so on when transferred to the full tree-sitter-sql since `field` no longer contains an internal `object_reference`

I think it is promising or at least a hopeful sign though! I'll keep looking into this

dmfay:
okay I'm alive! Lots of progress here -- actually the object_reference move is complete, and the remaining issue I can see is with the various optimize statements added in the interim. As pushed this doesn't compile, complaining about conflicts first in `vacuum` and then in others.

The `vacuum` conflict is resolved with `prec.right` but that's a very blunt instrument and there's something extra-fishy going on: one of the conflict paths involves `field _vacuum_table_repeat1`, although paren_list uses plain `repeat` not `repeat1`, and then if you change line 1490 to specify `identifier` the conflict now implicates `create_view_repeat1` which has a `paren_list($.identifier)`! Possibly we need to lift those into named/anonymous rules?

matthias-Q:
I don't understand why we have conflicts here at all. I mean `vacuum` is close to the root of the tree and it is followed by a node with is almost at the bottom (`field`; `identifier`).
From my understanding, this should not lead to any disambiguities as it is a fairly unique parsing "path".

What do you mean by "lifting into named/anonymous rules"?

dmfay:
> I don't understand why we have conflicts here at all. I mean `vacuum` is close to the root of the tree and it is followed by a node with is almost at the bottom (`field`; `identifier`). From my understanding, this should not lead to any disambiguities as it is a fairly unique parsing "path".
>
> What do you mean by "lifting into named/anonymous rules"?

because the anomalous `repeat1`s reference other rules entirely, I wonder if tree-sitter is doing something strange with scopes of "shared" anonymous rules like `paren_list` -- that is, because a `paren_list(identifier)` appears in the "create view" path first or last, other instances of `paren_list(identifier)` get mixed up with "create view" syntax. So all's well if we have only the one instance of `paren_list(identifier)` but the moment we have another we need to have an explicit `view_params_paren_list` + `other_thing_paren_list`, or at least that's my current guess.

dmfay:
I think it's because of how we structure programs as sequential statements and often have optional `;` delimiters. So `vacuum tbl (NEXT_TOKEN)` could express either a single statement expressing vacuum options _or_ two statements, a simplest-possible vacuum and a parenthesized expression.

That seems like a bit much to try to tackle here so I've added a conflict for _vacuum_table and _compute_stats which lets it compile for now. All tests pass, but we should revisit `program`s as sequences of delimited statements soon!

grammar.js
```diff
@@ -1327,14 +1353,16 @@ module.exports = grammar({
...
       optional($.table_partition), // Spark SQL
       optional(
         seq(
           $.keyword_as,
-          field('table_alias', $._alias_identifier)
+          field('alias', $.identifier)
         ),
       ),
+      // TODO i don't think we have a test for `insert into tbl set ....` and
+      // it's no dialect i can think of
```
matthias-Q, line 1365: If I read the docs correclty, then MariaDB allows for this:
https://mariadb.com/kb/en/insert/

grammar.js
```diff
@@ -2092,26 +2101,29 @@ module.exports = grammar({
       ),
       optional($.where),
       optional($.group_by),
+      optional($.window_clause),
```
matthias-Q, line 2104: is is correct that the window_clause can be between group_by and order_by? Does the order matter here? I thought that is should be last.

grammar.js
```diff
@@ -2092,26 +2101,29 @@ module.exports = grammar({
...
-        seq(
-          '(',
-          $.values,
-          ')',
+    relation: $ => prec.right(
+      seq(
+        choice(
+          $.subquery,
+          $.invocation,
+          // TODO already has an optional alias
```
matthias-Q, line 2114: whats the TODO here?

matthias-Q (approved):
This is awesome. I really like that we have a clearer structure now.

There 3 comments, two of them are just about comments in the code.

grammar.js
dmfay, line 1365:
```suggestion
      // TODO we need a test for `insert...set`
```

grammar.js
dmfay: I think this one's leftover from when `table_reference` included an alias as well as there being the alias here

```suggestion
```

grammar.js
dmfay: It does make sense as `limit` and `order by` apply to the result of set operations (union, intersect, except) & you can see those are a dividing line

PR #156, opened 2023-06-19
fix: distinct and order by in invocation

### LINKED ISSUE #154: `ERROR` on `COUNT(DISTINCT tbl.field)`
image

Tree-sitter syntax error for `tbl` and `field` in:

```sql
SELECT
  COUNT(DISTINCT `tbl`.`field`)
FROM
  whatever;
```

matthias-Q:
Closes #154 

Distinct has been misidentified as a field. The pg specific `order by` was not working at all (we had a test for that)

@DerekStride should we add a CI test that checks for `ERROR` in the tests? (excluding error.txt) to avoid this from happening again?

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1940,7 +1940,15 @@ module.exports = grammar({
 
     invocation: $ => seq(
       field('name', $.identifier),
-      paren_list(field('parameter', $._select_expression)),
+      paren_list(
+        seq(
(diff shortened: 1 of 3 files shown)


DerekStride (approved):
A CI check for `ERROR` nodes is a good idea 👍

PR #159, opened 2023-06-20
fix: `group_concat` on impala

### LINKED ISSUE #158: `group_concat` as generic `invocation`
In 8c0ca863 `group_concat` was introduced. However, this function exists in other dialects was well with a different internal structure (aka the parameters of the list)

e.g.
```sql
select
    group_concat(col1, ',')
from tab
```

This statement is not parsed correctly. I think `group_concat` should a generic node of type `invocation`

Alternatively, implement the complete internal expression syntax. It looks quite similar between MariaDB, MySQL and Impala. In that example above, the optional separator is missing.

matthias-Q:
`group_concat` is slightly different on impala. Instead of the keyword `separator` the sep char is written after a comma.

I am not happy with the fact that we have invocation and this second line of _aggregate_functions. Imo they should be also invocations but might be difficult to add every variant of `invocations`

Closes #158

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1928,7 +1928,10 @@ module.exports = grammar({
       field('name', $.keyword_group_concat),
       '(',
       $._aggregate_expression,
-      optional(seq($.keyword_separator, alias($._literal_string, $.literal))),
+      optional(seq(
+        choice($.keyword_separator, ','),
+        alias($._literal_string, $.literal)
+      )),
       ')',
     ),
 
--- a/test/corpus/functions.txt
+++ b/test/corpus/functions.txt
@@ -218,6 +218,38 @@ GROUP BY some_field;
           (identifier))))))
 
 ================================================================================
+GROUP CONCAT Impala
+================================================================================
+
+SELECT GROUP_CONCAT(uid, ",")
+FROM some_table
+GROUP BY some_field;
+
+--------------------------------------------------------------------------------
+
+(program
+  (statement
+    (select
+      (keyword_select)
+      (select_expression
+        (term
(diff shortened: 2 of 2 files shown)


matthias-Q:
Ping @DerekStride @dmfay What do you think regarding `group_concat`?

dmfay (approved):
this is fine but I do wonder whether it should be an `invocation` -- it looks for example a lot like Postgres' `string_agg`. Is there a reason it needs its own node type?

matthias-Q:
It has exactly the same purpose as `string_agg`. I also raised this question in the linked issue. I don't think it should be a separate thing.

`group_concat`in MySQL and MariaDB seem to have their own style and are not just simple invocations (parameter list is different) That's why is was implemented separately.

dmfay:
what about adding an aggregate path in `invocation`? this conflicts with `term` atm but I think we could make it work

```js
    invocation: $ => seq(
      field('name', $.identifier),
      choice(
        paren_list(
          seq(
            optional($.keyword_distinct),
            field(
              'parameter',
              $._select_expression,
            ),
            optional($.order_by)
          )),
        seq(
          '(',
          optional($.keyword_distinct),
          field('parameter', choice($._expression, $.all_fields)),
          optional($.order_by),
          optional(seq(
            choice($.keyword_separator, ','),
            alias($._literal_string, $.literal)
          )),
          ')',
        ),
      ),
    ),
```

matthias-Q:
That looks good. I just had a similar idea.

matthias-Q:
Closing this in favour of #163

PR #160, opened 2023-06-21
feat(tests): Add highlight tests

DerekStride:
## What

See https://github.com/nvim-treesitter/nvim-treesitter/pull/4979

Add new highlight tests to make integration with editors easier. The `highlights.scm` file can be copied as is to `nvim-treesitter`


matthias-Q (approved):
LGTM

PR #162, opened 2023-06-22
fix: impala partitioning variation

### LINKED ISSUE #161: `PARTITIONED BY` column_definitions for impala/hive is not correct
```sql
CREATE TABLE tab
PARTITIONED BY (col1, col2) -- <-- ERROR
STORED AS PARQUET AS
SELECT
    col1,
    col2,
    col3
FROM tab_2
```
The columns definitions list is the same as for a create table statement, however for the partitioning it is not necessary to add the datatype.

matthias-Q:
Impala/Hive have different variations of CREATE TABLE. Partitioning has a slightly different style for external and managed tables:

https://impala.apache.org/docs/build/html/topics/impala_create_table.html

For managed tables you do not have to specify the data type of the partition.

Closes #161

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1628,8 +1628,9 @@ module.exports = grammar({
         $.keyword_partition,
       ),
       choice(
-        seq('(', $.identifier, ')'), // postgres
-        $.column_definitions, // impala/hive
+        // seq('(', $.identifier, ')'), // postgres
+        seq('(', $.identifier, repeat(seq(',', $.identifier)), ')'), // postgres & Impala (CTAS)
+        $.column_definitions, // impala/hive external tables
         seq('(', $._key_value_pair, repeat(seq(',', $._key_value_pair)), ')',), // Spark SQL
       )
     ),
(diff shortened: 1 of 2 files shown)


grammar.js
```diff
@@ -1628,8 +1628,9 @@ module.exports = grammar({
         $.keyword_partition,
       ),
       choice(
-        seq('(', $.identifier, ')'), // postgres
-        $.column_definitions, // impala/hive
+        // seq('(', $.identifier, ')'), // postgres
+        seq('(', $.identifier, repeat(seq(',', $.identifier)), ')'), // postgres & Impala (CTAS)
```
dmfay, line 1632: you could use `comma_list` for this possibly but I'm still not sure what was going on with those in conflicts in #150 . Also the comment on line 1631 can be cleared

dmfay (approved):

grammar.js
matthias-Q: I can also use `paren_list`. I forgot that they exist :-) Thanks for the hint.

PR #163, opened 2023-06-23
refactor: `group_concat` and `_aggregate_function` to `invocation`

### LINKED ISSUE #158: `group_concat` as generic `invocation`

matthias-Q:
Add `_aggregate_function` as part of `invocation`.

Closes #158 and supersedes #159

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1937,17 +1937,15 @@ module.exports = grammar({
         $.object_reference,
         choice(
           // default invocation
-          seq(
-            paren_list(
-              seq(
-                optional($.keyword_distinct),
-                field(
-                  'parameter',
-                  $.term,
-                ),
-                optional($.order_by)
-              )
-            ),
+          paren_list(
+            seq(
+              optional($.keyword_distinct),
+              field(
+                'parameter',
+                $.term,
+              ),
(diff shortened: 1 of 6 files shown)


DerekStride (approved):
Thanks for pushing this forward 💯

grammar.js
```diff
@@ -1933,36 +1932,36 @@ module.exports = grammar({
...
         $.object_reference,
-        paren_list(
+        choice(
+          // default invocation
           seq(
-            optional($.keyword_distinct),
-            field(
-              'parameter',
-              $.term,
+            paren_list(
```
dmfay, line 1941: the seq on line 1940 is superfluous with this

dmfay (approved):

PR #164, opened 2023-06-24
refactor: programs as delimited sequences of statements

dmfay:
eliminates the recently added conflicts, shaves a few seconds off compile time, and state count in the parser is cut by over half -- from 7151 states to 3087!

Other changes:

- `compound_statement` has been renamed to `block`
- `create_function` uses `object_reference` correctly
- we've lost parsing of 'string' function bodies but that was probably a bad idea to begin with

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -11,12 +11,6 @@ module.exports = grammar({
   conflicts: $ => [
     [$.object_reference, $._qualified_field],
     [$.object_reference],
-    // TODO these two have internal conflicts because of optional parenthesized
-    // settings which can be interpreted as _subsequent statements_ due to our
-    // current handling of statement delimiters in program nodes. Remove once
-    // we have expressed programs as delimited sequences of statements.
-    [$._vacuum_table],
-    [$._compute_stats],
   ],
 
   precedences: $ => [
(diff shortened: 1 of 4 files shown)


grammar.js
```diff
@@ -489,13 +492,48 @@ module.exports = grammar({
     // https://stackoverflow.com/questions/13014947/regex-to-match-a-c-style-multiline-comment
     marginalia: _ => seq('/*', /[^*]*\*+(?:[^/*][^*]*\*+)*/, '/' ),
 
-    compound_statement: $ => seq(
```
matthias-Q, line 492: In nvim-treesitter is a indentens.scm which refenrences `compound_Statement` we should have added a copy of this to our repo as we so that the tests fail here when we rename nodes.

matthias-Q (approved):
Wow, I am really surprised hat these few change have such a big impact. I am really impressed! Thanks @dmfay

matthias-Q:
> we've lost parsing of 'string' function bodies but that was probably a bad idea to begin with
what statements are you referring to?

dmfay:
> > we've lost parsing of 'string' function bodies but that was probably a bad idea to begin with
>
> what statements are you referring to? Can you give me an example?

there are a couple in the tests, for example

```
create or replace function public.fn()
  returns int
  language sql
as 'select 1;';
```

`select 1;` is now parsed as a `literal` instead of a `statement` in its own right. It could be a `program` if we could ensure that any single quotes internal to it didn't parse in the _outer_ program, which is probably possible but a cursory check didn't look promising.

DerekStride (approved):
Wow this is great, the decrease in number of states is huge 🎉🎉🎉

Issue #166, opened 2023-06-28
Parsing PostgreSQL JSON operators


Gelio:
I've noticed [PostgreSQL JSON operators](https://www.postgresql.org/docs/9.3/functions-json.html) are not parsed correctly.

For example, the following query has parsing errors:

```sql
SELECT "user"->>'login' AS username FROM users;
```

Parsed AST:

```
(program
  (statement
    (select
      (keyword_select)
      (select_expression
        (term
          (binary_expression
            (literal)
            (ERROR)
            (literal))
          (keyword_as)
          (identifier))))
    (from
      (keyword_from)
      (relation
        (object_reference
          (identifier))))))
```

matthias-Q:
JSON ist not implemented yet. Thanks for reporting. This will probably be a bigger undertaking.

Issue #128, opened 2023-03-29
`NOT IN` precedence issues


matthias-Q:
As discussed in #126 `NOT IN` is currently not implemented

DerekStride:
## Additional context

Implementing `NOT IN` as mentioned in #126, is causing some errors with precedence. I think the root cause might be from the between expression added in #117. It has a rule `seq($._expression, $.keyword_and, $._expression)` but the `binary_expression` node is also defined as `seq($._expression, $.keyword_and, $._expression)`.

Some investigation & work is required to detangle the precedence issues and implement `NOT IN`.

Issue #115, opened 2023-03-03
Postgres Array function is not parsed correctly


matthias-Q:
```sql
select array(select 'x')
```

Is a valid SQL statement in Postgres. The odd thing is, that it does not seems to be documented.

Since we have the `array` node for statements like this
```sql
select array['x']
```

This actually gives the same result.

PR #170, opened 2023-06-29
feat: json traversal

dmfay:
this needs more time than I have for it at the moment but I don't think it'll be too hard to get basic support for #166! element indexing (possibly some common ground with the `array[1,2,3]` syntax?) and jsonpath are bigger considerations though.


matthias-Q:
I added the other json operators and updated the test. Actually it looks like this is it. All other json operations are probably handled by `invocation`

Gelio:
Drive-by appreciation comment:

I am honestly amazed by your aptitude and the promptness at fixing the issues I've reported. Stellar project maintenance! I certainly wasn't expecting to see all 3 issues addressed the same day I reported them

dmfay:
> Drive-by appreciation comment:
>
> I am honestly amazed by your aptitude and the promptness at fixing the issues I've reported. Stellar project maintenance! I certainly wasn't expecting to see all 3 issues addressed the same day I reported them

the JSON issue has secretly been bugging me for a while (along with array literals and subscripting, but that's for later) and I've always been too busy with other things & other projects to take a look, and it turns out it just fits right into the `expression` node now

matthias-Q (approved):
LGTM. Thanks for fixing the tests.

test/corpus/json.txt
```diff
@@ -0,0 +1,47 @@
...
+   (select_expression
+    (term
+     (json_traversal
+      (literal))
+     (keyword_as)
+     (identifier))
+    (term
+     (json_traversal
+      (json_traversal
+       (literal))))
```
DerekStride, line 28: This subtree represents the term `"user"->'address' ->> 'city'`. It only references a single `literal` node which means the other two have been hidden.

I've puts together https://github.com/DerekStride/tree-sitter-sql/pull/171 to show an alternative using the `binary_expression` node that will surface all 3 literals and treat the navigation nodes as anonymous operators.

matthias-Q:
We should add the new operators to the highlights.scm (i forgot that with `filter` yesterday as well ...)

matthias-Q:
This looks good to me. I am merging it. I am so surprised that this went so smoothly.

(I was hoping that after the recent refactor we could also tackle `array` and `not in`, but it seems like they are still difficult to solve)

dmfay:
> This looks good to me. I am merging it. I am so surprised that this went so smoothly.
>
> (I was hoping that after the recent refactor we could also tackle `array` and `not in`, but it seems like they are still difficult to solve)

I went looking and we [actually got arrays](https://github.com/DerekStride/tree-sitter-sql/blob/main/test/corpus/select.txt#L2398) at some point!

matthias-Q:
@dmfay yes, but not as a function. See #115 Besides, I think #128 is more important :-/

PR #172, opened 2023-06-29
Moving bindings and package.swift to work with Swift

bombardier200:
Since moving the src folder to a different branch this has broken the Package.swift and therefore no longer works for swift. By moving the package.swift and bindings folder to the gh-pages branch, it will now work with SPM and allow people to use it in Swift. 
If these changes are merged then the package.swift could be removed from main and I assume also the bindings folder could be removed.
I have included all bindings here as I know it has to be here for swift to work but unsure about rust and node.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,0 +1,33 @@
+// swift-tools-version:5.3
+
+import PackageDescription
+
+let package = Package(
+    name: "TreeSitterSQL",
+    platforms: [.macOS(.v10_13), .iOS(.v11)],
+    products: [
+        .library(name: "TreeSitterSQL", targets: ["TreeSitterSQL"]),
+    ],
+    dependencies: [],
+    targets: [
+        .target(name: "TreeSitterSQL",
+                path: ".",
+                exclude: [
+                    "binding.gyp",
+                    "bindings",
+                    "Cargo.toml",
+                    "grammar.js",
+                    "LICENSE",
+                    "Makefile",
+                    "package.json",
+                    "README.md",
+                    "src/grammar.json",
+                    "src/node-types.json",
+                ],
+                sources: [
+                    "src/parser.c",
+                ],
+                publicHeadersPath: "bindings/swift",
+                cSettings: [.headerSearchPath("src")])
+    ]
+)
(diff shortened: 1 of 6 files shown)


DerekStride:
This won't work unfortunately, the `gh-pages` branch is automatically updated by this [workflow](https://github.com/DerekStride/tree-sitter-sql/blob/main/.github/workflows/gh-pages.yml) everytime a new commit is merged into `main`. This PR will just get overridden on the next merge to the `main` branch.

Is it possible to run `tree-sitter generate` or, alternatively, `npm install` after cloning this repository to make sure all the parser artifacts are generated?

bombardier200:
https://github.com/marketplace/actions/github-pages-action#%EF%B8%8F-keeping-existing-files-keep_files 
if you added this flag to the action I would think it would keep those files.

mattmassicotte:
Perhaps ironically, it was also difficult to add SPM support to the [Swift parser](https://github.com/alex-pinkus/tree-sitter-swift) for similar reasons. The solution there was to modify the workflow to copy the needed files. Would you be open to that instead?

mattmassicotte:
> This won't work unfortunately, the `gh-pages` branch is automatically updated by this [workflow](https://github.com/DerekStride/tree-sitter-sql/blob/main/.github/workflows/gh-pages.yml) everytime a new commit is merged into `main`. This PR will just get overridden on the next merge to the `main` branch.
> 
> Is it possible to run `tree-sitter generate` or, alternatively, `npm install` after cloning this repository to make sure all the parser artifacts are generated?

This is possible for some environments, but requiring a functional tree-sitter cli or node install raises the bar high enough that it would appear not to work for most clients. If that's required, I think we'd probably just maintain a fork.

DerekStride:
> The solution there was to modify the workflow to copy the needed files. Would you be open to that instead?

Yeah that's a good option too, I think this would be all that's required to copy the files over in the workflow

```diff
- run: cp -r src docs 
+ run: cp -r src bindings Package.swift docs 
```

https://github.com/DerekStride/tree-sitter-sql/blob/318b0e4d4fbd2a9417e1b90d1a96261d2db1fe39/.github/workflows/gh-pages.yml#L19

bombardier200:
Yes that seems like all it would take.

DerekStride:
fixed in #177

thecoolwinter:
I don't mean to resurrect this issue but would you be open to adding the `queries` folder to that branch? I'd be happy to make a PR to add it rq if you're okay with that addition.

PR #175, opened 2023-07-03
feat: tsql multiple add columns in alter statement

### LINKED ISSUE #135: T-SQL Alter Table

matthias-Q:
Closes #135 and supersedes #144

I feel this is a bit hacky, but I just made the `add` keyword optional, so that it will not repeated in `_alter_specifications`

I also removed that `rename` part of the test, since this is not valid SQL syntax and made the keyword `constrain` optional. See discussion in #144

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1131,6 +1131,8 @@ module.exports = grammar({
       $.change_ownership,
     ),
 
+    // TODO: optional `keyword_add` is necessary to allow for chained alter statements in t-sql
+    // maybe needs refactoring
     add_column: $ => seq(
       optional($.keyword_add),
       optional(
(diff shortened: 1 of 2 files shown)


DerekStride (approved):

dmfay (approved):
it's a venial kludge at worst imo but a TODO comment above add_column might be a good idea in case we wind up revisiting

Issue #176, opened 2023-07-04
unions after cte aren't parsed properly


Appalled:
Dear tree-sitter-sql developer, 
Unions  after cte aren't parsed properly, here are two examples. Please have a look. 
```
with tb2 as (
  SELECT * FROM tb1
)
(
  (SELECT * FROM tb2)
  UNION
  (SELECT * FROM tb2)
)
;

with tb2 as (
  SELECT * FROM tb1
)
(SELECT * FROM tb2)
UNION
(SELECT * FROM tb2)
;

```

matthias-Q:
Thanks for reporting. The issues are the brackets, not the unions. Will take a closer look tomorrow.

Appalled:
> Thanks for reporting. The issues are the brackets, not the unions. Will take a closer look tomorrow.

Thanks for your swift response. The forked nvim-treesitter I use is not always in sync with the upstream, so my feedback maybe a little laggy. Previous examples got fixed, while I found another problem. Please have a look:
```
SELECT 
a1.*
FROM (
  SELECT * FROM tb01
  UNION ALL
  SELECT * FROM tb01
) a1
;
```

PR #178, opened 2023-07-04
Update unions with named node & allow more optional paranthesis

DerekStride:
## What

fixes https://github.com/DerekStride/tree-sitter-sql/issues/176

Commits might be easier to review separately

* New `optional_parenthesis` & `wrapped_in_parenthesis` functions to make it easier work with parenthesis
* Alias custom `create_query` node with `_dml_read`
* New ~`union` & `intersection`~ `set_operation` node instead of having custom rules in `_select_statement`

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -550,13 +550,8 @@ module.exports = grammar({
     statement: $ => seq(
       choice(
         $._ddl_statement,
-        $._dml_read,
         $._dml_write,
-        seq(
-          '(',
-          $._dml_read,
-          ')',
-        ),
+        optional_parenthesis($._dml_read),
       ),
     ),
 
@@ -1700,15 +1695,8 @@ module.exports = grammar({
     ),
 
     _default_expression: $ => seq(
-        $.keyword_default,
-            choice(
-            seq(
-                '(',
-                    $._inner_default_expression,
-                ')',
-            ),
-            $._inner_default_expression,
-        )
+      $.keyword_default,
+      optional_parenthesis($._inner_default_expression),
     ),
     _inner_default_expression: $ => choice(
         $.literal,
--- a/grammar.js
+++ b/grammar.js
@@ -593,12 +593,11 @@ module.exports = grammar({
     ),
 
     _dml_read: $ => seq(
-      choice(
-        seq(
-          optional(
-            $._cte
-          ),
+      optional($._cte),
+      optional_parenthesis(
+        choice(
           $._select_statement,
+          $.set_operation,
         ),
       ),
     ),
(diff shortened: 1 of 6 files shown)


grammar.js
```diff
@@ -584,12 +579,12 @@ module.exports = grammar({
...
-        seq(
-          optional(
-            $._cte
-          ),
+      optional($._cte),
+      optional_parenthesis(
+        choice(
           $._select_statement,
+          $.union,
+          $.intersection,
```
dmfay, line 587: there's also an `except` set operation that yields all tuples in a not present in b

grammar.js
```diff
@@ -541,13 +541,8 @@ module.exports = grammar({
...
       choice(
         $._ddl_statement,
-        $._dml_read,
         $._dml_write,
-        seq(
-          '(',
-          $._dml_read,
-          ')',
-        ),
+        optional_parenthesis($._dml_read),
```
dmfay, line 545: okay it's really weird that you can wrap a `select` but not an `insert` or a `delete`

test/corpus/cte.txt
```diff
@@ -231,52 +231,52 @@ FROM top_cte;
 --------------------------------------------------------------------------------
 
 (program
- (statement
-  (keyword_with)
-  (cte
-   (identifier)
-   (keyword_as)
-   (statement
+  (statement
```
dmfay, line 234: missing a possible test for cte in parentheses -- this parses in Postgres:

```sql
((((with x as ((((select * from ints)))) ((((select * from x))))))));
```

grammar.js
```diff
@@ -2485,6 +2459,15 @@ function unsigned_type($, type) {
   )
 }
 
+function optional_parenthesis(node) {
+  return prec.right(
```
dmfay, line 2463: does it not work without right-precedence?

matthias-Q (approved):
Wow, many changes. LGTM, I hope nothing breaks :smile: 

Maybe add this extra function that I suggested in a comment.

grammar.js
matthias-Q: @dmfay did you tested that? It is really astonishing how many things are not documented in postgres

grammar.js
DerekStride: I had merged that with `intersection` should these all be grouped under a common `set_operation` node instead?

grammar.js
DerekStride: Nope, I think it's because there's multiple layers where the parenthesis are allowed to appear so this pushes it "down" the tree.

```
$ tree-sitter generate
Unresolved conflict for symbol sequence:

  '('  union  •  ')'  …

Possible interpretations:

  1:  '('  (_dml_read  union)  •  ')'  …
  2:  (_dml_read  '('  union  •  ')')

Possible resolutions:

  1:  Specify a left or right associativity in `_dml_read`
  2:  Add a conflict for these rules: `_dml_read`
```

test/corpus/cte.txt
DerekStride: I added a test for multiple layers of parenthesis but an arbitrary number of parenthesis might require a custom scanner.

DerekStride:
I added a few more commit to address some PR comments I also have an open question here: https://github.com/DerekStride/tree-sitter-sql/pull/178#discussion_r1252268536

I'd appreciate another round of reviews 🎉

grammar.js
dmfay: oh I like having the generic node type discriminable by the `operation` field in the third!

grammar.js
dmfay: yeah even if you use a `returning` it refuses to parse any parenthesized write statement :exploding_head:

test/corpus/cte.txt
dmfay: dollar quotes (beyond the few common types I added explicitly) need a custom scanner too, but I didn't have much luck with that when I tried a while back

grammar.js
dmfay: ah weird, that's probably it. Not a lot to do about it I suppose

matthias-Q (approved):
So many changes. Since it is now mostly using the new functions I would so go for it. looks good!

dmfay (approved):

Issue #155, opened 2023-06-19
`keyword_language` is unused, renderet to `parser.c` as `aux_sym_*`


LeoniePhiline:
While preparing https://github.com/helix-editor/helix/pull/7387 I noticed that a `(keyword_language)` query makes tree-sitter fail parsing with a `NodeType` error.

Upon investigation, I found that:

[The symbol](https://github.com/DerekStride/tree-sitter-sql/blob/92018a36833c284c3f5d923de7ec253f35c6e610/grammar.js#L201) [is defined](https://github.com/DerekStride/tree-sitter-sql/blob/gh-pages/src/grammar.json#L666-L669) in `grammar.json`.

But it is not defined in [`node-types.json`](https://github.com/DerekStride/tree-sitter-sql/blob/gh-pages/src/node-types.json).

In [`parser.c`](https://raw.githubusercontent.com/DerekStride/tree-sitter-sql/gh-pages/src/parser.c), it is defined as `aux_sym_keyword_language_token1`. 🤔 (see also [the codegen](https://github.com/tree-sitter/tree-sitter/blob/d0029a15273e526925a764033e9b7f18f96a7ce5/cli/src/generate/render.rs#L1482))

`(keyword_language)` [appears to be unused](https://github.com/DerekStride/tree-sitter-sql/blob/92018a36833c284c3f5d923de7ec253f35c6e610/grammar.js#L946).

I think it can be removed - or shall it be used for `function_language`?

PR #179, opened 2023-07-04
fix: removed unused keyword

DerekStride:
## What

fixes https://github.com/DerekStride/tree-sitter-sql/issues/155

Use the `keyword_language` node in the grammar rules.

### DIFF
--- a/queries/highlights.scm
+++ b/queries/highlights.scm
@@ -206,6 +206,8 @@
   (keyword_optimize)
   (keyword_vacuum)
   (keyword_cache)
+  (keyword_language)
+  (keyword_sql)
 ] @keyword
 
 [
--- a/grammar.js
+++ b/grammar.js
@@ -967,7 +967,7 @@ module.exports = grammar({
     ),
 
     function_language: $ => seq(
-      make_keyword('language'),
+      $.keyword_language,
       choice(
         $.keyword_sql,
         $.keyword_plpgsql,
--- a/test/corpus/functions.txt
+++ b/test/corpus/functions.txt
@@ -483,6 +483,7 @@ return 1;
       (int
         (keyword_int))
       (function_language
+        (keyword_language)
         (keyword_sql))
       (function_body
         (keyword_return)
@@ -521,6 +522,7 @@ return 1;
       (int
         (keyword_int))
       (function_language
+        (keyword_language)
         (keyword_sql))
       (function_body
         (keyword_return)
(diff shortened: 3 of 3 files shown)


matthias-Q (approved):

DerekStride:
@matthias-Q I'm requesting another review, I added some more commits that ensure our `keyword_` nodes are in sync with the `highlights.scm` file.

You can see the failure message here: https://github.com/DerekStride/tree-sitter-sql/actions/runs/5456777435/jobs/9930044879

And see it fixed in the latest commit.

matthias-Q:
I am bit short on time today. I can take a look tomorrow. I was noticing a lot of "new" keywords in the highlights. Did we missed all of them? Do we need to add them to nvim-treesitter as well? If so, I would maybe put some of them not under the `@keyword` query.

DerekStride:
>  Did we missed all of them? Do we need to add them to nvim-treesitter as well?

yeah they were all missing & need to be synced to nvim-treesitter.

> I would maybe put some of them not under the @keyword query.

I moved a few of them to `@attribute`, there's still a lot of generic `@keywords` but I'm not too sure where else they could belong

queries/highlights.scm
```diff
@@ -191,21 +198,44 @@
...
   (keyword_cached)
   (keyword_uncached)
   (keyword_lines)
   (keyword_stored)
-  (keyword_location)
   (keyword_partitioned)
-  (keyword_cached)
   (keyword_analyze)
   (keyword_rewrite)
-  (keyword_statistics)
```
matthias-Q, line 205: I think `statistics` and `stats` should be `@type.qualifier` and not `@keyword`

matthias-Q (approved):
Looks good. I made two suggestions for the highlights, but in the end I expect the maintainers of nvim-treesitter to have their own opinion :smile:

queries/highlights.scm
```diff
@@ -9,7 +9,6 @@
   (keyword_spgist)
   (keyword_gin)
   (keyword_brin)
-  (keyword_float)
```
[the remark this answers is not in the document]
DerekStride: We already have `float` in `@type.builtin` and the rule for `invocation` should cover the `@function.call` case.

queries/highlights.scm
DerekStride: I removed these from here because they were duplicated in `@type.qualifier` already 👍

PR #180, opened 2023-07-10
fix: subquery unions & invalid tests

### LINKED ISSUE #176: unions after cte aren't parsed properly

matthias-Q:
Fixes #176

I had to remove the optional `;` from the `subquery` node. I think that this was wrong. It does not seem to be allowed in Postgres.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -2372,11 +2372,7 @@ module.exports = grammar({
     ),
 
     subquery: $ => wrapped_in_parenthesis(
-      seq(
-        $.select,
-        optional($.from),
-        optional(";"),
-      ),
+      $._dml_read
     ),
 
     list: $ => paren_list($._expression),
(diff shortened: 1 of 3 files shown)


DerekStride (approved):

grammar.js
```diff
@@ -2372,11 +2372,7 @@ module.exports = grammar({
     ),
 
     subquery: $ => wrapped_in_parenthesis(
-      seq(
-        $.select,
-        optional($.from),
-        optional(";"),
-      ),
+      $._dml_read
```
DerekStride, line 2375: Nice and simple, the AST is really benefitting from the `dml_read` nodes 🎉

PR #182, opened 2023-07-14
feat: adds `merge into` upsert statements

matthias-Q:
This PR add the `MERGE INTO` statement used by Trino/Athena/Snowflake/Iceberg 

Reference:
https://trino.io/docs/current/sql/merge.html
https://docs.snowflake.com/en/sql-reference/sql/merge
https://docs.aws.amazon.com/athena/latest/ug/merge-into-statement.html

Note: 
I also added the "keyword_check" bash script to the test command, so that we check it during development and not just during CI

I think T-SQL and Oracle also support this statement. On a quick glance, Oracle should be covered, T-SQL has some extra commands.

[Update]: I just realized that Postgres also supports this :astonished: It also is almost fully covered! https://www.postgresql.org/docs/current/sql-merge.html


grammar.js
```diff
@@ -1381,6 +1384,47 @@ module.exports = grammar({
       optional($.returning),
     ),
 
+    _upsert_statement: $=> seq(
```
dmfay, line 1387: I don't think any dialect actually uses a keyword `upsert` but imo it'd still be clearer to call this `_merge_statement`

test/corpus/insert.txt
```diff
@@ -526,3 +526,228 @@ FROM
                     (identifier))
                   (literal)))))
           (identifier))))))
+
```
dmfay, line 529: we've only got two tests for `merge` so far but if we expand them we'll want to have a dedicated merge.txt so why not start now?

grammar.js
```diff
@@ -1381,6 +1384,47 @@ module.exports = grammar({
...
+      repeat1($.when_clause)
+    ),
+
+    when_clause: $ => seq(
+      $.keyword_when,
+      optional($.keyword_not),
+      $.keyword_matched,
+      optional(
+        seq($.
+          keyword_and,
```
dmfay, line 1409:
```suggestion
        seq(
          $.keyword_and,
```

dmfay (approved):
couple little bookkeeping issues but lgtm!

PR #185, opened 2023-07-27
feat(pg): create columns as array and matrices

### LINKED ISSUE #183: PG: Create Table with Array types
Using the `[]` notation in PostgresSQL, it is possible to create columns that store arrays and matrices.

```sql
CREATE table tab (
    arr_col INT[],
    mat_col INT[][]
);
```

This is currently not handled.

matthias-Q:
Closes #183

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -384,61 +384,77 @@ module.exports = grammar({
 
     keyword_array: _ => make_keyword("array"), // not included in _type since it's a constructor literal
 
-    _type: $ => choice(
-      $.keyword_boolean,
-      $.bit,
-      $.keyword_binary,
+    _type: $ => seq(
+      choice(
+        $.keyword_boolean,
+        $.bit,
+        $.keyword_binary,
 
-      $.keyword_smallserial,
-      $.keyword_serial,
-      $.keyword_bigserial,
+        $.keyword_smallserial,
+        $.keyword_serial,
+        $.keyword_bigserial,
 
-      $.tinyint,
-      $.smallint,
-      $.mediumint,
-      $.int,
-      $.bigint,
-      $.decimal,
-      $.numeric,
-      $.double,
-      $.float,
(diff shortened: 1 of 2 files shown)


grammar.js
```diff
@@ -384,61 +384,77 @@ module.exports = grammar({
...
+      choice(
+        seq(optional($.keyword_array), $._array_size_definition),
+        seq( $._array_size_definition, $._array_size_definition),
+        $.keyword_array,
+      ),
+    ),
+    _array_size_definition: $ => seq(
+      '[',
+      optional(field("size", alias($._integer, $.literal))),
+      ']'
```
dmfay, line 457:
```suggestion
      seq(
        optional($.keyword_array),
        repeat(
          seq(
            '[',
            optional(field("size", alias($._integer, $.literal))),
            ']'
          )
        )
      ),
```

Postgres allows n-dimensional arrays > 2; `repeat` is 0 or more times so this should cover all possibilities

matthias-Q: the ARRAY keyword works only for 1D arrays
shouldn't it be `repeat1()` instead of `repeat()`?

matthias-Q: I have pushed a version, which allows for more dimensions. Unfortunately, it requires a conflict rule.

dmfay: `repeat` would allow `array`, `array[n]`, `array[n][n]....` and cover all three at once (including the zero-repeats case of just the keyword), but it is more permissive than necessary if the keyword is only used with 1d arrays. I'm fine either way though.

if you replace the top and bottom options in your latest with `seq($.keyword_array, optional($._array_size_definition))` do you still need the conflict rule?

matthias-Q: Ah yes, this resolves the conflict.

I have pushed it to my branch, but for some reason it does not show up in the PR :confused:

dmfay (approved):

PR #186, opened 2023-07-29
Custom type handling

### LINKED ISSUE #184: PG: Support creation and usage of  custom types
Postgres allows to create custom types. The creation of types, tables that are using these types and the selection of columns with these types is not yet supported:

Here is an example:

```sql
CREATE TYPE obj AS (
    elem1 TEXT,
    elem2 INT
);


CREATE TABLE obj_tab (
    object obj
);

SELECT
    (object).elem1,
    (object).elem2,
FROM obj_tab
```

Docs:
https://www.postgresql.org/docs/current/sql-createtype.html

matthias-Q:
This PR adds handling of custom types in Postgres:

* `create type`
  * `enum`
  * `range`
  * object types
* `drop type`
* `alter type`
  * `rename`, `rename attributes`
  * `drop`, `add`, `alter` attributes
  * ownership & schema
* `create table` with a column as a custom type
* select from a table with a custom attribute (dot notation)

Remarks:
We have many statement with a `set ( KEY = VALUE)` section. The keys are often specific, but as of now we allow everything. They could be technically treated as keywords, but we have to add a lot of different variants. Or we keep it as it is and the parser will be more permissive.
(I think I asked this question before, but I don't remember the outcome of the discussion...)

References:
https://www.postgresql.org/docs/current/sql-createtype.html
https://www.postgresql.org/docs/current/sql-altertype.html
https://www.postgresql.org/docs/current/sql-droptype.html

Closes #184


test/corpus/custom_types.txt
```diff
@@ -0,0 +1,443 @@
...
+  (statement
+    (create_type
+      (keyword_create)
+      (keyword_type)
+      (identifier)
+      (keyword_as)
+      (keyword_enum)
+      (literal)
+      (literal)
+      (literal))))
```
dmfay, line 38: seeing this inline I wonder if it'd be useful to have a containing expression for the values -- is filtering for `literal`s specific enough?

grammar.js
```diff
@@ -1241,11 +1290,79 @@ module.exports = grammar({
...
+          $.identifier,
+          $.keyword_to,
+          $.identifier,
+          optional($._cascade_restrict)
+        ),
+        seq(
+          $.keyword_add,
+          $.keyword_value,
+          optional($._if_not_exists),
+          $.identifier,
```
dmfay, line 1313: this should be `literal`, identifiers cannot be specified as enum values (even trying to get cute with it in a procedure)

```sql
[local] dian#dian= create procedure test(newval text) language plpgsql as $$ begin
[local] dian#dian$ alter type boxes add value if not exists newval;
[local] dian#dian$ return;
[local] dian#dian$ end;
[local] dian#dian$ $$;
ERROR:  syntax error at or near "newval"
LINE 2: alter type boxes add value if not exists newval;
                                                 ^
Time: 23.887 ms
```

grammar.js
```diff
@@ -1241,11 +1290,79 @@ module.exports = grammar({
...
+          optional(
+            seq(
+              choice($.keyword_before, $.keyword_after),
+              $.identifier
```
dmfay, line 1317:
```suggestion
              $.literal
```

same here

test/corpus/custom_types.txt
```diff
@@ -0,0 +1,443 @@
...
+        (keyword_set)
+        (keyword_schema)
+        (identifier)))))
+
+================================================================================
+ALTER TYPE ADD VALUE
+================================================================================
+
+ALTER TYPE boxes
+ADD VALUE  IF NOT EXISTS  color  AFTER weight
```
dmfay, line 233:
```suggestion
ADD VALUE IF NOT EXISTS 'color' AFTER weight
```

test/corpus/custom_types.txt
```diff
@@ -0,0 +1,443 @@
...
+      (keyword_type)
+      (identifier)
+      (keyword_add)
+      (keyword_value)
+      (keyword_if)
+      (keyword_not)
+      (keyword_exists)
+      (identifier)
+      (keyword_after)
+      (identifier))))
```
dmfay, line 250:
```suggestion
      (literal)
      (keyword_after)
      (literal))))
```

test/corpus/custom_types.txt
```diff
@@ -0,0 +1,443 @@
...
+(program
+  (statement
+    (alter_type
+      (keyword_alter)
+      (keyword_type)
+      (identifier)
+      (set_schema
+        (keyword_set)
+        (keyword_schema)
+        (identifier)))))
```
dmfay, line 226:
```suggestion
```
this duplicates the previous test

test/corpus/custom_types.txt
```diff
@@ -0,0 +1,443 @@
...
+      (identifier)
+      (keyword_after)
+      (identifier))))
+
+================================================================================
+ALTER TYPE RENAME VALUE
+================================================================================
+
+ALTER TYPE boxes
+RENAME VALUE weight TO mass
```
dmfay, line 257:
```suggestion
RENAME VALUE 'weight' TO 'mass'
```

test/corpus/custom_types.txt
```diff
@@ -0,0 +1,443 @@
...
+  (statement
+    (alter_type
+      (keyword_alter)
+      (keyword_type)
+      (identifier)
+      (keyword_rename)
+      (keyword_value)
+      (identifier)
+      (keyword_to)
+      (identifier))))
```
dmfay, line 271:
```suggestion
      (literal)
      (keyword_to)
      (literal))))
```

grammar.js
```diff
@@ -1241,11 +1290,79 @@ module.exports = grammar({
...
+            )
+          ),
+        ),
+        seq(
+          $.keyword_rename,
+          $.keyword_value,
+          $.identifier,
+          $.keyword_to,
+          $.identifier,
```
dmfay, line 1326:
```suggestion
          $.literal,
          $.keyword_to,
          $.literal,
```
same, this is modifying enum values not composite type attributes

dmfay (changes requested):
> We have many statement with a set ( KEY = VALUE) section. The keys are often specific, but as of now we allow everything. They could be technically treated as keywords, but we have to add a lot of different variants. Or we keep it as it is and the parser will be more permissive.
> (I think I asked this question before, but I don't remember the outcome of the discussion...)

a lot of these get really dialect-specific so imo it's fine to be permissive with them

test/corpus/custom_types.txt
matthias-Q: There are no field identifiers shown in the test. If you parse it with tree-sitter directly, this looks like this:

```
(program [0, 0] - [5, 0]
  (statement [0, 0] - [4, 1]
    (create_type [0, 0] - [4, 1]
      (keyword_create [0, 0] - [0, 6])
      (keyword_type [0, 7] - [0, 11])
      (identifier [0, 12] - [0, 18])
      (keyword_as [0, 19] - [0, 21])
      (keyword_enum [0, 22] - [0, 26])
      enum_element: (literal [1, 4] - [1, 9])
      enum_element: (literal [2, 4] - [2, 11])
      enum_element: (literal [3, 4] - [3, 10]))))
```

I will make it an extra node, so it will be clearer.

matthias-Q:
I have change from `identifier` to `literal` for add/rename values of a custom type. It is now correct. However, it will also parse it as correct, when these fields are wrapped in double quotes. I think only single quotes will work. Maybe we should split the `_literal_string` in two hidden nodes so that we can used them separately.

Also, on my testing postgres instance this does not work:

```sql
ALTER TYPE color RENAME VALUE 'purple' TO 'mauve';
```
It is taken directly from the docs and it complaints about the keyword value. Am I missing something very obvious here?

dmfay:
> I have change from `identifier` to `literal` for add/rename values of a custom type. It is now correct. However, it will also parse it as correct, when these fields are wrapped in double quotes. I think only single quotes will work. Maybe we should split the `_literal_string` in two hidden nodes so that we can used them separately.
>
> Also, on my testing postgres instance this does not work:
>
> ```sql
> ALTER TYPE color RENAME VALUE 'purple' TO 'mauve';
> ```
>
> It is taken directly from the docs and it complaints about the keyword value. Am I missing something very obvious here?

is the color type still an enum or did you recreate it as a composite while testing things?

```
[local] dian#dian= create type color as enum ('purple', 'red', 'yellow');
CREATE TYPE
Time: 14.847 ms
[local] dian#dian= alter type color rename value 'purple' to 'mauve';
ALTER TYPE
Time: 14.992 ms
```

dmfay (approved):

matthias-Q:
I just had a super old postgres version. I have updated to 15.3 and now it works.

PR #187, opened 2023-07-30
Add `explain`, `truncate` and `alter`/`drop`/`create schema`

matthias-Q:
This PR adds multiple missing statements 
* for schemas 
* truncate
* explain

I have left out the optional options for explain. They add multiple pg specific keywords. Not sure how we want to handle this. Also, statements that can be analyzed are a mixture of some of the `_dml_read`, `_dml_write` and `_ddl_statement`. It is more permissive now, but it also required this precedence. Unfortunately, this significantly increases compilation time and parser size again.

`CREATE SCHEMA` allows to be followed by other CREATE statements. Again, it is now permissive and requires a precedence.


grammar.js
```diff
@@ -1056,6 +1071,23 @@ module.exports = grammar({
...
+          optional($._if_not_exists),
+          $.identifier,
+          optional(seq($.keyword_authorization, $.identifier)),
+        ),
+        seq(
+          $.keyword_authorization,
+          $.identifier,
+        ),
+      ),
+      repeat($._create_statement),
```
dmfay, line 1088: structuring this form into `program` (now a `choice`) and deleting the repeat line seems to work well. It does require precedence over _create_statement -> create_schema but I think that's reasonable

```js
      prec(2, seq(
        $.create_schema,
        repeat($._create_statement),
      )),
```

grammar.js
```diff
@@ -570,6 +574,7 @@ module.exports = grammar({
     ),
 
     statement: $ => seq(
+      optional($._explain_statement),
```
dmfay, line 577:
```suggestion
      optional(seq(
        $.keyword_explain,
        optional($.keyword_analyze),
        optional($.keyword_verbose),
      )),
```
`explain` won't be reused and we can cut out the precedence

dmfay (changes requested):

grammar.js
matthias-Q: Not sure that I understand this correctly. Isn't the `program` node a bit too close to the root of the tree?

I have put it in the `create_statement` with a left prec.

dmfay (approved):
> Not sure that I understand this correctly. Isn't the `program` node a bit too close to the root of the tree?
>
> I have put it in the `create_statement` with a left prec.

oh that's a good place for it!

PR #188, opened 2023-08-01
Missing CREATE, ALTER DROP statement

matthias-Q:
This PR adds `CREATE`, `ALTER`, `DROP` for:
* `sequence`
* `database`
* `user`, `group`, `role`
* `index` (only `alter`)

This is quite a lot but very repetitive. Unfortunately, many times it is just 80% reusable code and many dialect specific keyword. I have tried to find a balance.

I noticed that we still miss a lot of basic statements. I will add an issue for that. it would have been to much to put it in here as well.


grammar.js
```diff
@@ -1094,6 +1115,94 @@ module.exports = grammar({
...
+    ),
+
+    create_database: $ => seq(
+      $.keyword_create,
+      $.keyword_database,
+      $.identifier,
+      optional($.keyword_with),
+      repeat(
+        $._with_settings
+      ),
```
dmfay, line 1131:
```suggestion
      optional(
        seq(
          $.keyword_with,
          repeat1(
            $._with_settings
          ),
        ),
      ),
```
prevent settings without `with` -- I think this goes for some of the other additions too?

grammar.js
```diff
@@ -1094,6 +1115,94 @@ module.exports = grammar({
...
+        optional($.keyword_encrypted),
+        $.keyword_password,
+        choice(
+          field("password", alias($._literal_string, $.literal)),
+          $.keyword_null,
+        ),
+      ),
+    ),
+
+    _user_access_role_config: $ => prec.left(seq(
```
dmfay, line 1173: did you break this out of _role_options for the precedence? you should be able to apply `prec.left` to an individual element in a `choice` -- I did it the other day testing the `create schema` stuff

grammar.js
```diff
@@ -1339,6 +1448,146 @@ module.exports = grammar({
...
+              choice(
+                $.keyword_all,
+                field("option", $.identifier),
+              )),
+          ),
+        )
+      ),
+    ),
+
+    set_configuration: $ => seq(
```
dmfay, line 1515: I think this could be consolidated further -- not all the way up to `keyword_set` unfortunately but including the field (picking one of option/configuration_parameter) and the 'from current'

grammar.js
```diff
@@ -1339,6 +1448,146 @@ module.exports = grammar({
...
+            optional(seq($.keyword_cache, field("cache", alias($._integer, $.literal)))),
+            optional(seq(optional($.keyword_no), $.keyword_cycle))
+          )),
+          optional(seq(
+            $.keyword_restart,
+            optional($.keyword_with),
+            field("restart", alias($._integer, $.literal)),
+          )),
+          optional(seq($.keyword_owned, $.keyword_by, choice($.keyword_none, $.object_reference))),
+        ),
```
dmfay, line 1578: this is suspect -- there are a lot of options in the docs but the `start` sequence is repeated, and it's not possible to `alter sequence x no cycle` etc because those options are locked into the `start` sequence. I think something like this is the best result, a seq of optional seqs in the [documentation order](https://www.postgresql.org/docs/current/sql-altersequence.html):

```
seq(
  optional(seq($.keyword_as), <type>),
  optional(seq($.keyword_increment, optional($.keyword_by), $.literal)),
  optional(seq($.keyword_minvalue, choice($.literal, 'no minvalue')),
  optional(seq($.keyword_maxvalue, choice($.literal, 'no maxvalue')),
  optional(seq($.keyword_start, optional($.keyword_with), $.literal)),
  etc
)
```

grammar.js
```diff
@@ -1440,6 +1692,26 @@ module.exports = grammar({
       optional($._drop_behavior)
     ),
 
+    drop_database: $ => seq(
+      $.keyword_drop,
+      $.keyword_database,
+      optional($._if_exists),
+      $.identifier,
+      optional($.keyword_with),
+      optional($.keyword_force),
```
dmfay, line 1701: it's really `with force`! oh wow I had no idea, that's great

test/corpus/alter.txt
```diff
@@ -697,3 +697,292 @@ ALTER SCHEMA sales OWNER TO CURRENT_USER;
...
+          (identifier))))))
+
+================================================================================
+ALTER INDEX
+================================================================================
+
+ALTER INDEX distributors SET (fillfactor = 75);
+ALTER INDEX distributors RESET (fillfactor);
+ALTER INDEX coord_idx ALTER COLUMN 3 SET STATISTICS 1000;
+ALTER INDEX distributors SET TABLESPACE fasttablespace;
```
dmfay, line 730: these individual checks are good but a quick "maximal" test setting as many options as possible would be useful for each of the complex object specifications here. Probably more manageable as individual cases than rolled in like these are

dmfay (commented):
thanks for getting these started!

grammar.js
matthias-Q: I broke it out, because I was unhappy with more spaghetti code :-) I just realized that the prec.left is not necessary. I will remove it.

grammar.js
matthias-Q: I think this is wrong: 
https://www.postgresql.org/docs/current/sql-createdatabase.html

I think the `with` is optional and then there can be 0 or more `_with_options`
```
CREATE DATABASE name
    [ WITH ] [ OWNER [=] user_name ]
           [ TEMPLATE [=] template ]
...
```

grammar.js
matthias-Q: Oh, there are some copy and paste mishaps. I have fixed that, added the missing variants and I accept my questionable formatting :-)

test/corpus/alter.txt
matthias-Q: I have separated them a little bit more, so that similar tests are together. I also added some more "complex" tests.

grammar.js
```diff
@@ -1094,6 +1117,96 @@ module.exports = grammar({
...
+      optional(seq($.keyword_increment, optional($.keyword_by), field("increment", alias($._integer, $.literal)))),
+      optional(seq($.keyword_minvalue, choice($.literal, seq($.keyword_no, $.keyword_minvalue)))),
+      optional(seq($.keyword_maxvalue, choice($.literal, seq($.keyword_no, $.keyword_maxvalue)))),
+      optional(seq(
+        $.keyword_start,
+        optional($.keyword_with),
+        field("start", alias($._integer, $.literal)),
+        optional(seq($.keyword_cache, field("cache", alias($._integer, $.literal)))),
+        optional(seq(optional($.keyword_no), $.keyword_cycle))
+      )),
```
matthias-Q, line 1206: This section is shared with `alter sequence``
I cannot break this out because they are all optionals and tree-sitter cannot match on potentially empty nodes. Any ideas?

dmfay: cache and cycle clauses are distinct from start! the `cache` and `cycle` keywords respectively are required

dmfay: oh whoops I misidentified what you were talking about, the clauses are correct in alter but still combined here. As for consolidating with `alter` I'm not sure that's practical unfortunately -- the alter version has a choice with additional options (ownership, renaming) that don't apply

matthias-Q: I was trying to break out the parts that are shared among create and alter sequence:
```
_subset: $ => seq(
  optional(seq($.keyword_as, $._type)),
  optional(seq($.keyword_increment, optional($.keyword_by), $.literal)),
  optional(seq($.keyword_minvalue, choice($.literal, seq($.keyword_no, $.keyword_minvalue)))),
  optional(seq($.keyword_maxvalue, choice($.literal, seq($.keyword_no, $.keyword_maxvalue)))),
  optional(seq($.keyword_start, optional($.keyword_with), field("start", alias($._integer, $.literal)))),
),
```

but this will not work, because it can match on an empty string. So, we could also keep it as it is. I mean this subset will only be used in by two statements. I could live with that.

dmfay: yeah I think the only way you could separate an anonymous node is if you captured the $.object_reference too, but then alter has other options and you lose the `choice`

you might be able to break it out into a _function_ sequence_options() that returns the all-optional `seq` though!

grammar.js
```diff
@@ -1339,6 +1452,129 @@ module.exports = grammar({
...
+      $.object_reference,
+      choice(
+        seq(
+          optional(seq($.keyword_as, $._type)),
+          optional(seq($.keyword_increment, optional($.keyword_by), $.literal)),
+          optional(seq($.keyword_minvalue, choice($.literal, seq($.keyword_no, $.keyword_minvalue)))),
+          optional(seq($.keyword_maxvalue, choice($.literal, seq($.keyword_no, $.keyword_maxvalue)))),
+          optional(seq($.keyword_start, optional($.keyword_with), field("start", alias($._integer, $.literal)))),
+          optional(seq($.keyword_restart, optional($.keyword_with), field("restart", alias($._integer, $.literal)))),
+          optional(seq($.keyword_cache, field("cache", alias($._integer, $.literal)), optional(seq(optional($.keyword_no), $.keyword_cycle)))),
```
dmfay, line 1563: cache and cycle are separate (`alter sequence x no cycle` is allowed), otherwise this is the correct version of this block -- the "create" version still has both cache and cycle baked into `start`

grammar.js
matthias-Q: I think I keep it as it is. A function for a section that will only occur twice does not sound good.

grammar.js
matthias-Q: Ah it o weird that in the documentation, they put these optional sections in one line. That is a bit confusing.

I have fixed that.

dmfay: the `(no) cycle` path is having a really outsized effect on build time and state count -- just commenting those out in create/alter sequence brings us down from 8313 states to 6143, and build times back to 50s on my machine. But all those optionals do a lot: if you comment out the `as type` in `create sequence` as well we're down to 5787 states, comment it out in `alter sequence` instead 6021 (I expected `alter` to be heavier since it's more deeply nested, weird!).

But it seems Postgres isn't too particular about the exact order: `alter sequence s no cycle restart with 1` is accepted. So we can do this instead:

```javascript
    create_sequence: $ => seq(
      $.keyword_create,
      optional(
        choice(
          choice($.keyword_temporary, $.keyword_temp),
          $.keyword_unlogged,
        )
      ),
      $.keyword_sequence,
      optional($._if_not_exists),
      $.object_reference,
      repeat1(
        choice(
          seq($.keyword_as, $._type),
          seq($.keyword_increment, optional($.keyword_by), field("increment", alias($._integer, $.literal))),
          seq($.keyword_minvalue, choice($.literal, seq($.keyword_no, $.keyword_minvalue))),
          seq($.keyword_maxvalue, choice($.literal, seq($.keyword_no, $.keyword_maxvalue))),
          seq($.keyword_start, optional($.keyword_with), field("start", alias($._integer, $.literal))),
          seq($.keyword_cache, field("cache", alias($._integer, $.literal))),
          seq(optional($.keyword_no), $.keyword_cycle),
          seq($.keyword_owned, $.keyword_by, choice($.keyword_none, $.object_reference)),
        ),
      ),
    ),

    alter_sequence: $ => seq(
      $.keyword_alter,
      $.keyword_sequence,
      optional($._if_exists),
      $.object_reference,
      choice(
        repeat1(
          choice(
            seq($.keyword_as, $._type),
            seq($.keyword_increment, optional($.keyword_by), $.literal),
            seq($.keyword_minvalue, choice($.literal, seq($.keyword_no, $.keyword_minvalue))),
            seq($.keyword_maxvalue, choice($.literal, seq($.keyword_no, $.keyword_maxvalue))),
            seq($.keyword_start, optional($.keyword_with), field("start", alias($._integer, $.literal))),
            seq($.keyword_restart, optional($.keyword_with), field("restart", alias($._integer, $.literal))),
            seq($.keyword_cache, field("cache", alias($._integer, $.literal))),
            seq(optional($.keyword_no), $.keyword_cycle),
            seq($.keyword_owned, $.keyword_by, choice($.keyword_none, $.object_reference)),
          ),
        ),
        $.rename_object,
        $.change_ownership,
        seq(
          $.keyword_set,
          choice(
            choice($.keyword_logged, $.keyword_unlogged),
            seq($.keyword_schema, $.identifier)
          ),
        ),
      ),
    ),
```

and wind up with 4275 states and 36s build time

matthias-Q: Thanks, that explains a lot.

dmfay (approved):

PR #196, opened 2023-09-26
feat: add natural join

antoineB:


grammar.js
```diff
@@ -2669,6 +2671,11 @@ module.exports = grammar({
       )
     ),
 
+    natural_join: $ => seq(
```
matthias-Q, line 2674: you could also add `natural` as another choice in the `join` node which includes already `left`, `righter`... and so on instead of creating a new node `natural_join`

matthias-Q (changes requested):
Hi, thanks for the PR. I made just one comment. I don't see a reason why `natural_join` should exists alongside the normal `join` node.

grammar.js
```diff
@@ -75,6 +75,7 @@ module.exports = grammar({
     keyword_cross: _ => make_keyword("cross"),
     keyword_join: _ => make_keyword("join"),
     keyword_lateral: _ => make_keyword("lateral"),
+    keyword_natural: _ => make_keyword("natural"),
```
matthias-Q, line 78: please add this keyword also to the hightlights.scm

matthias-Q: under `@keyword` would be a good place (next to `keyword_lateral` for example)

grammar.js
antoineB: Done.

grammar.js
antoineB: Done.

matthias-Q (approved):

matthias-Q:
Thanks for the MR @antoineB

PR #197, opened 2023-09-26
Feature/on delete

antoineB:


grammar.js
```diff
@@ -75,6 +75,7 @@ module.exports = grammar({
     keyword_cross: _ => make_keyword("cross"),
     keyword_join: _ => make_keyword("join"),
     keyword_lateral: _ => make_keyword("lateral"),
+    keyword_natural: _ => make_keyword("natural"),
```
matthias-Q, line 78: this is already part of #196 right?

grammar.js
```diff
@@ -2669,6 +2690,11 @@ module.exports = grammar({
       )
     ),
 
+    natural_join: $ => seq(
```
matthias-Q, line 2693: This is already part of #196

grammar.js
```diff
@@ -2581,6 +2601,7 @@ module.exports = grammar({
       repeat(
         choice(
           $.join,
+          $.natural_join,
```
matthias-Q, line 2604: This is already part of #196

grammar.js
```diff
@@ -2273,13 +2275,31 @@ module.exports = grammar({
             seq(
               $.keyword_on,
               $.keyword_delete,
-              $.keyword_cascade,
+              $._referential_action,
```
matthias-Q, line 2278: this could be combined in one optional block with a choice between `keyword_delete` and `keyword_update`. I see no difference between the other parts of the sequence.

matthias-Q (changes requested):
Thanks for the PR. I made some remarks that need to be addressed.

grammar.js
```diff
@@ -234,6 +235,7 @@ module.exports = grammar({
     keyword_nowait: _ => make_keyword("nowait"),
     keyword_attribute: _ => make_keyword("attribute"),
     keyword_authorization: _ => make_keyword("authorization"),
+    keyword_action: _ => make_keyword("action"),
```
matthias-Q, line 238: this new keyword also needs to be added in the highlights.scm

matthias-Q: maybe as a `@type.qualifier` ?

grammar.js
antoineB: Yes I used my previous MR to avoid conflicts.

grammar.js
antoineB: Yes.

grammar.js
antoineB: I put it into keyword.

grammar.js
```diff
@@ -2269,11 +2271,20 @@ module.exports = grammar({
...
-              $.keyword_delete,
-              $.keyword_cascade,
+              choice($.keyword_delete, $.keyword_update),
+              choice(
+                seq($.keyword_no, $.keyword_action),
+                $.keyword_restrict,
+                $.keyword_cascade,
+                seq(
+                  $.keyword_set,
+                  choice($.keyword_null, $.keyword_default),
```
matthias-Q, line 2284: I think `set null` and `set default` (this one is missing), can optionally be followed by a column name. In our parser the column name is represented by the `object_reference` node. (`paren_list()` of `object_reference`

https://www.postgresql.org/docs/current/sql-altertable.html (under `referential_action`)

matthias-Q (changes requested):
Thanks for the changes. I double checked the statement in the Postgres documentation and it looks like a small part is missing. Would be nice if you could add this.

matthias-Q:
@antoineB Thanks for the MR. I have two general remarks:
Would be nice if you would not `force push` changes that are requested in a review. It makes it easier to review. We can always squash commits before merging. It would also be nice if you could add a small query that is failing or not yet supported and also provide the link to the SQL Dialect documentation. This makes it a lot easier for us. 

However, we should definitely think about some contribution rules

grammar.js
antoineB: You cannot put can just put an object_reference just an identifier:
```
toto=# create table test1(id integer references test0(id) on delete set null(test.id));
ERROR:  syntax error at or near "."
LINE 1: ...d integer references test0(id) on delete set null(test.id));
```

antoineB: Also line 2273 `$.ordered_columns` should be `paren_list($.identifier)` at least on postgres.

matthias-Q: This worked for me on postgres 15.3

```sql
CREATE TABLE table1 (
    id serial PRIMARY KEY,
    column_name1 integer,
    column_name2 integer
);

CREATE TABLE table2 (
    id serial PRIMARY KEY,
    table1_id integer REFERENCES table1(id) ON DELETE CASCADE
);

ALTER TABLE table1
DROP CONSTRAINT IF EXISTS table1_column_name1_fkey;

ALTER TABLE table1
ADD CONSTRAINT table1_column_name1_fkey
FOREIGN KEY (column_name1)
REFERENCES table2(id)
ON DELETE SET NULL (column_name1);
```

matthias-Q: You are right about `object_reference`. This won't work. `identifier` does work however.

matthias-Q (approved):

Issue #189, opened 2023-08-01
Missing Statments


matthias-Q:
Following statement types are still missing:

- [ ] `GRANT`
- [ ] `REVOKE`
- [ ] `COMMENT`

https://www.postgresql.org/docs/15/sql-grant.html
https://www.postgresql.org/docs/15/sql-revoke.html
https://www.postgresql.org/docs/15/sql-comment.html

PR #199, opened 2023-10-03
Add comment statement

ds-cbo:
Part of #189

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -724,8 +724,6 @@ module.exports = grammar({
       ),
     ),
 
-    comment_text: $ => $._literal_string,
-
     comment_statement: $ => seq(
       $.keyword_comment,
       $.keyword_on,
@@ -733,7 +731,7 @@ module.exports = grammar({
       $.keyword_is,
       choice(
         $.keyword_null,
-        $.comment_text,
+        alias($._literal_string, $.literal),
       ),
     ),
 
--- a/queries/highlights.scm
+++ b/queries/highlights.scm
@@ -31,7 +31,6 @@
 
 (literal) @string
 (comment) @comment @spell
-(comment_text) @string
 (marginalia) @comment
 
 ((literal) @number
--- a/test/corpus/comment_stat.txt
+++ b/test/corpus/comment_stat.txt
@@ -1,5 +1,5 @@
 ================================================================================
-Comment table
+Comment on table
 ================================================================================
 
 COMMENT ON TABLE my_schema.my_table IS "this table is a test";
@@ -16,13 +16,13 @@ COMMENT ON TABLE my_schema.my_table IS "this table is a test";
 				schema: (identifier)
 		    	name: (identifier))
 			(keyword_is)
-			(comment_text))))
+			(literal))))
 
 ================================================================================
-Comment column is null
+Comment on column is null
 ================================================================================
 
-COMMENT ON COLUMN my_table.my_column IS NULL;
+COMMENT ON COLUMN my_schema.my_table.my_column IS NULL;
 
 --------------------------------------------------------------------------------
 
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -1,19 +1,39 @@
-ts := $(shell which tree-sitter 2> /dev/null)
-ifeq (, ${ts})
-	ts := $(shell which tree-sitter-cli 2> /dev/null)
+CC ?= gcc
+
+TS := $(shell which tree-sitter 2>/dev/null)
+ifeq (, ${TS})
+	TS := $(shell which tree-sitter-cli 2>/dev/null)
 endif
+TSFLAGS ?=
```
ds-cbo, line 7: for example: `make test TSFLAGS=--debug-build` to speed up the testing cycle

grammar.js
```diff
@@ -723,6 +724,59 @@ module.exports = grammar({
...
+        $.comment_text,
+      ),
+    ),
+
+    _comment_target: $ => choice(
+      // TODO: access method
+      // TODO: aggregate
+      $.cast,
+      // TODO: collation
+      seq($.keyword_column, $.object_reference),
```
dmfay, line 745:
```suggestion
      seq($.keyword_column, alias($._qualified_field, $.object_reference)),
```
`comment on column schema.table.column is 'asdf';` wouldn't parse since `object_reference` only covers up to two parts. `_qualified_field` technically allows a parenthesized schema+table but I don't think we need to add a whole new node type for tripartite names over this

grammar.js
```diff
@@ -723,6 +724,59 @@ module.exports = grammar({
...
+      // TODO: function
+      seq($.keyword_index, $.object_reference),
+      // TODO: large object
+      seq($.keyword_materialized, $.keyword_view, $.object_reference),
+      // TODO: operator (|class|family)
+      // TODO: policy
+      // TODO: (procedural) language
+      // TODO: procedure
+      // TODO: publication
+      seq($.keyword_role, $.object_reference),
```
dmfay, line 763:
```suggestion
      seq($.keyword_role, $.identifier),
```

grammar.js
```diff
@@ -723,6 +724,59 @@ module.exports = grammar({
...
+      // TODO: routine
+      // TODO: rule
+      seq($.keyword_schema, $.object_reference),
```
dmfay, line 766:
```suggestion
      seq($.keyword_schema, $.identifier),
```
schema can only be the one name

grammar.js
```diff
@@ -723,6 +724,59 @@ module.exports = grammar({
...
+      seq($.keyword_sequence, $.object_reference),
+      // TODO: server
+      // TODO: statistics
+      // TODO: subscription
+      seq($.keyword_table, $.object_reference),
+      seq($.keyword_tablespace, $.object_reference),
```
dmfay, line 772:
```suggestion
      seq($.keyword_tablespace, $.identifier),
```

grammar.js
```diff
@@ -723,6 +724,59 @@ module.exports = grammar({
...
+      // TODO: text search (configuration|dictionary|parser|template)
+      // TODO: transform for
+      seq($.keyword_trigger, $.keyword_on, $.object_reference),
```
dmfay, line 775:
```suggestion
      seq($.keyword_trigger, $.identifier, $.keyword_on, $.object_reference),
```
`comment on trigger` needs a (unipart) name

grammar.js
```diff
@@ -723,6 +724,59 @@ module.exports = grammar({
       ),
     ),
 
+    comment_text: $ => $._literal_string,
```
dmfay, line 727: comment text is a reasonably special case imo but consider the common `alias($._literal_string, $.literal)` pattern here!

dmfay (commented):
thanks for the makefile improvements! there are a few incorrect paths in `_comment_target`; it's pretty self-contained so I'm not too too worried about it but a little more test coverage would also help

grammar.js
ds-cbo: Ah, I was already wondering how to get a literal string to be visible in the test cases, I hadn't figured out `alias` yet. Thanks!

dmfay (approved):

ds-cbo:
Thanks!

PR #205, opened 2023-10-15
Fix ALTER TABLE ... ADD CONSTRAINT ... UNIQUE

antoineB:
Make the statement work with postgres syntax while keeping the previous mariadb syntax (which seems to be used in create table only).

Why does `ALTER TABLE "Role" ADD CONSTRAINT "pkRole" PRIMARY KEY ("roleId");` CST has `ordered_columns` while
`ALTER TABLE table_name ADD CONSTRAINT constraint_name UNIQUE (col1, col2, col3);` does not ?

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -2324,7 +2324,7 @@ module.exports = grammar({
         $.keyword_index,
       ),
       optional(field('name', $.identifier)),
-      paren_list($.identifier, true),
+      $.ordered_columns,
       optional(
         seq(
           $.keyword_references,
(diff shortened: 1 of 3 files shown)


dmfay (approved):
> Why does `ALTER TABLE "Role" ADD CONSTRAINT "pkRole" PRIMARY KEY ("roleId");` CST has `ordered_columns` while `ALTER TABLE table_name ADD CONSTRAINT constraint_name UNIQUE (col1, col2, col3);` does not ?

no idea; afaik both of those are SQL Server-isms, and I haven't used that in a decade and change

PR #211, opened 2023-10-28
feat: Add missing (postgres) operators.

antoineB:
The operators of posix regexp, bit string, network addresses, point, full text search, range and some missing json operators have been added.

Following postgres operators precedence
https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-PRECEDENCE. They have the 'binary_other' precedence rule, and to reduce the size of the parser they all produce the either token 'op_other' or 'op_unary_other'.


matthias-Q (approved):
LGTM. I think it is time to update downstream projects like nvim-treesitter, helix etc. Their highlights.scm queries should know the newly added keywords and tokens.

PR #215, opened 2023-10-28
feat: Add more string litral formats (from postgres) and numbers foramts

antoineB:
Add bit string literal
https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-BIT-STRINGS Add unicode string
https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS-UESCAPE Add exponent in integers and decimals
https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-CONSTANTS-NUMERIC Add hexadecimal, binary and octal formats
Add '_' as separator of digits
Add string casting (eg: REAL '1.23')


grammar.js
```diff
@@ -3124,8 +3126,18 @@ module.exports = grammar({
...
+      /(0[xX][0-9A-Fa-f]+(_[0-9A-Fa-f]+)*)|(0[oO][0-7]+(_[0-7]+)*)|(0[bB][01]+(_[01]+)*)|(\d+(_\d+)*(e[+-]?\d+(_\d+)*)?)/
+    ),
+    _decimal_number: $ => seq(
+      optional(
+        choice("-", "+")),
+      /((\d+(_\d+)*)?[.]\d+(_\d+)*(e[+-]?\d+(_\d+)*)?)|(\d+(_\d+)*[.](e[+-]?\d+(_\d+)*)?)/
+    ),
+    _bit_string: $ => seq(/[bBxX]'([^']|'')*'/, repeat(/'([^']|'')*'/)),
+    // The identifier should be followed by a string (no parenthesis allowed)
+    _string_casting: $ => seq($.identifier, $._single_quote_string),
```
matthias-Q, line 3140: This is essentially the same as `implicit_cast`. Could that be added there as an alternative?

matthias-Q (approved):
I have one question, otherwise LGTM.

grammar.js
antoineB: In spirit it is a literal and not a cast, it only work with a type followed by a literal string.
It is not a cast because I think it is done at "compile time".
```
postgres=# select REAL '1.25';
 float4 
--------
   1.25
(1 row)

postgres=# select REAL '' ;
ERROR:  invalid input syntax for type real: ""
LINE 1: select REAL '' ;
                    ^
postgres=# select REAL '' || '1.25';
ERROR:  invalid input syntax for type real: ""
LINE 1: select REAL '' || '1.25';
                    ^
postgres=# select REAL ('' || '1.25');
ERROR:  syntax error at or near "("
LINE 1: select REAL ('' || '1.25');
                    ^
postgres=# select REAL ('1.25');
ERROR:  syntax error at or near "("
LINE 1: select REAL ('1.25');
                    ^
postgres=# select REAL '1.25';
 float4 
--------
   1.25
(1 row)

postgres=# select REAL column1 from (values('1.25'));
ERROR:  column "real" does not exist
LINE 1: select REAL column1 from (values('1.25'));
               ^
```

Issue #209, opened 2023-10-26
Doubling of STATE_COUNT following dollar_quote feature merging


antoineB:
Now (5660d80) the state_count = 13456 roughly doubling, the size of the binary is 1.8Mo previously 1.2Mo.

Either

```diff
index 69d0a1d..2421aa2 100644
--- a/grammar.js
+++ b/grammar.js
@@ -1216,7 +1216,7 @@ module.exports = grammar({
         $.keyword_as,
         alias($._dollar_quoted_string_start_tag, $.dollar_quote),
         $._function_body_statement,
-        optional(';'),
+        ';',
         alias($._dollar_quoted_string_end_tag, $.dollar_quote),
       ),
     ),
```

Or

```diff
index 69d0a1d..e8faa61 100644
--- a/grammar.js
+++ b/grammar.js
@@ -1217,7 +1217,7 @@ module.exports = grammar({
         alias($._dollar_quoted_string_start_tag, $.dollar_quote),
         $._function_body_statement,
         optional(';'),
-        alias($._dollar_quoted_string_end_tag, $.dollar_quote),
+        $.dollar_quote,
       ),
     ),
```

Fix the explosion state_count I don't know why.

antoineB:
I try to have the closer to what we have but unfortunately it did work, I suppose it is because `$.dollar_quote` and `$._dollar_quoted_string_end_tag` lex the same chars.

```diff
index 69d0a1d..606f042 100644
--- a/grammar.js
+++ b/grammar.js
@@ -21,6 +21,7 @@ module.exports = grammar({
     [$.between_expression, $.binary_expression],
     [$.time],
     [$.timestamp],
+    [$._function_body_rule_a, $._function_body_rule_b],
   ],
 
   precedences: $ => [
@@ -1212,13 +1213,39 @@ module.exports = grammar({
           $.literal
         ),
       ),
-      seq(
-        $.keyword_as,
-        alias($._dollar_quoted_string_start_tag, $.dollar_quote),
-        $._function_body_statement,
-        optional(';'),
-        alias($._dollar_quoted_string_end_tag, $.dollar_quote),
-      ),
+      // This should be one rule but it is decomposed into 2 rules (A, B) to
+      // avoid state parser increase.
+      //
+      // It will parse valid SQL, but rule B will not fail if there is unbalanced
+      // dollar_quote or if there is another other dollar_quote inside
+      // $._function_body_statement with the same tag as the external dollar_quote.
+      //
[...]

PR #216, opened 2023-10-29
fix: Change the parser to require a semicolon in function_body of 1 s…

antoineB:
…tatement.

In order to reduce the number of parser state the grammar are less complient, now the function_body of 1 statement needs to end with a semicolon.

Idealy the parse of the function_body should be done by another parser in regard of the language defined.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1216,7 +1216,7 @@ module.exports = grammar({
         $.keyword_as,
         alias($._dollar_quoted_string_start_tag, $.dollar_quote),
         $._function_body_statement,
-        optional(';'),
+        ';',
         alias($._dollar_quoted_string_end_tag, $.dollar_quote),
       ),
     ),
--- a/test/corpus/functions.txt
+++ b/test/corpus/functions.txt
@@ -586,7 +586,7 @@ Function details, specified after string body
 
 create or replace function public.fn()
   returns int
-  as $$select 1$$
+  as $$select 1;$$
   language sql
   volatile
   parallel restricted


antoineB:
see https://github.com/DerekStride/tree-sitter-sql/issues/209

DerekStride:
If semi-colons are optional in this location I don't know if we should enforce them to simplify the parser. That said our parser is pretty large so I see why it's appealing. Are there any other thoughts?

Perhaps it's time to split the function body parsing into a separate parser?

dmfay:
yeah, the semicolon is really optional here and if we require it we will violate at least Postgres' expectations and I think the standard. It's worth looking imo at the state (as it were) of this parser relative to others. Here's a not very scientific sampling:

| parser | state count | large state count |
|--------|-------------|-------------------|
| Java | 1381 | 406 |
| JavaScript | 2781 | 440 |
| Fortran | 2921 | 447 |
| Rust | 3227 | 782 |
| LaTeX | 3371 | 2022 |
| Common Lisp | 4165 | 405 |
| SQLite (dhcmrlchtdj) | 5043 | 6 |
| Ruby | 5815 | 2122 |
| Bash | 6164 | 426 |
| Elixir | 6986 | 1091 |
| COBOL (Sakamoto) | 7744 | 5095 |
| C++ | 8043 | 2123 |
| **SQL (Stride)** | 14149 | 187 |
| Scala | 18538 | 2415 |
| SQL (Novikov) | 19852 | 4737 |
| Haskell | 22407 | 2777 |
| Kotlin | 25983 | 15612 |
| Julia | 31161 | 15000 |

We're above the median and mean, but it's hard to draw an exact parallel between SQL and other kinds of languages (COBOL might come the closest), _and_ we aim to support several SQL dialects. The main effect of high state counts is annoying development experience. We should try to keep it down but I don't think this is a crisis. Also worth noting our very low large state count compared to almost everything else.

Julia people are used to waiting forever for things to precompile ;)

matthias-Q:
I would suggest we close this PR. I think keeping it optional is the right choice.

PR #222, opened 2023-11-08
Add missing scanner.c to README and Package.swift

leamingrad:
Prior to this change, a new scanner.c source file had been added to the repository. The Makefile had been updated, but the command in the README was missing the new file.

This change adds the new file to the command in the README, and also updates the Package.swift file to declare it as a source.

I think this PR is just a bugfix following from #203.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -26,6 +26,7 @@ let package = Package(
                 ],
                 sources: [
                     "src/parser.c",
+                    "src/scanner.c",
                 ],
                 publicHeadersPath: "bindings/swift",
                 cSettings: [.headerSearchPath("src")])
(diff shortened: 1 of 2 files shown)


Package.swift
```diff
@@ -26,6 +26,7 @@ let package = Package(
                 ],
                 sources: [
                     "src/parser.c",
+                    "src/scanner.c",
```
leamingrad, line 29: I don't know anything about swift, so I am not sure whether this change is correct/how to test it (but happy to have a go if others can explain).

DerekStride: I'm unsure as well but I suspect this is all that's needed.

cc// @mattmassicotte let us know if there's anything else that's needed

DerekStride (approved):

Package.swift
mattmassicotte: This looks good - thanks for keeping it up to date!

PR #223, opened 2023-11-09
publish @derekstride/tree-sitter-sql npm package

### LINKED ISSUE #195: Publish on NPM
Hi, is it possible to publish this grammar on npm?

There is already a `tree-sitter-npm` package published (from this repository: https://github.com/m-novikov/tree-sitter-sql/), however it has not been updated in 2 years and this grammar seems to be more complete and active.

DerekStride:
fixes #195 

This sets up a workflow to publish to npm when PRs are merged into main.

I have a few open questions that anyone who has experience publishing to npm can help with:

1. Will we need to continually bump the version number in order to publish?
2. Should we instead use git tags to publish dedicated versions instead on every change?

cc// @rien


.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,19 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 21
+          registry-url: 'https://registry.npmjs.org'
+      - run: npm ci
+      - run: npm publish --access public.
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
DerekStride, line 19: I already setup this token ☑️

dmfay:
> 1. Will we need to continually bump the version number in order to publish?

yes, you can't republish the same version number

> 2. Should we instead use git tags to publish dedicated versions instead on every change?

I think this is a good idea! Especially because we don't have a way to control semver (there are things like [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) with npm plugins but they do require some discipline to use)

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,19 @@
...
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 21
+          registry-url: 'https://registry.npmjs.org'
+      - run: npm ci
+      - run: npm publish --access public.
```
rien, line 17: You could also use https://github.com/JS-DevTools/npm-publish, it only publishes a new version if the package version differs fro the last one on npm. We're using it for our npm package here: https://github.com/dodona-edu/dolos/blob/main/.github/workflows/ci-cd.yml#L529

But this command is much simpler, and might do the same check anyway.

DerekStride:
I've updated the PR to publish to npm on `publish` instead of during push to `main`. This will mean we need to publish an [new release](https://github.com/DerekStride/tree-sitter-sql/releases/new) on Github before it get's pushed to NPM.

@dmfay @matthias-Q I'm all for adopting [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) & [semantic versioning](https://semver.org/) and starting to track breaking changes to the schema if both of you are on board as well. We can unofficially start now but should start ourselves & asking other contributors to follow the guidelines after this PR lands and we publish our first release.

I've also added a `CONTRIBUTING.md` file & included a section on commit messages to encourage the use of [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) as suggested. The file also explains how to run tests, run the github pages server, and prepare a release. I'd appreciate feedback & suggestions on the document as I want it to reflect a consensus on our thoughts for the project.

CONTRIBUTING.md
```diff
@@ -0,0 +1,78 @@
...
+
+```
+$ cd docs/
+$ bundle install
+$ bundle exec jekyll serve
+```
+
+## Pushing a new Version
+
+Create a new PR to bump the version number in `package.json`. Bump the major version if the commit log since the last
```
dmfay, line 59: conventional commits let us automate this! add [commit-and-tag-version](https://www.npmjs.com/package/commit-and-tag-version) as a dev dependency, with a `release` npm script to run it as described. Then any collaborator can `npm run release` locally to generate the version bump commit (including a [changelog](https://gitlab.com/monstrous/monstrous/-/blob/main/CHANGELOG.md?ref_type=heads)) and tag it, and pushing main + tags will let the action take over

dmfay (approved):

DerekStride:
Any final comment or should I ship 🚢 this PR and publish our first release! 🎉

matthias-Q:
I say go for it. 
(Tip: I have set up a pre-commit hook in my machine to check for conventional commits)

DerekStride:
Published 🎉  https://www.npmjs.com/package/@derekstride/tree-sitter-sql

PR #224, opened 2023-11-09
Remove gitattributes

DerekStride:
In #100 we removed the generated files an publish them via GitHub pages. Since then we've added a custom src/scanner.c file that is being treated as a generated file on GitHub due to the .gitattributes file. Since it's no longer needed, I've removing it.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,0 @@
-src/** linguist-generated
-binding.gyp linguist-generated
-bindings/** linguist-generated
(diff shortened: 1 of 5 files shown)


dmfay (approved):

matthias-Q:
I am curious why the python import error occurs on the macos test runner?

DerekStride:
> I am curious why the python import error occurs on the macos test runner?

That got resolved when I updated the node version for some reason and there was another bug with the [nan](https://github.com/nodejs/nan/issues/942) package seemed to be fixed by updating the dependency.

PR #226, opened 2023-11-14
fix: segfault in scanner.c

### LINKED ISSUE #221: segfault
Hi 

somewhere between 44e36038f032eff2f36522642a8cc4351e133f83 ( good ) and 764c1bd5ea856dd7db5171783e442321dc3e041f (bad)  segfault when opening sql file.

unfortunately I can not post sql file here ( prod code )

I am not familiar enough with tree-sitter, whats the easiest way to run parser though any sql file ?

I've tried adding offending sql file to `test/highlight` but that did't do anything
```
syntax highlighting:
  ✓ union.sql (13 assertions)
  ✓ query.sql (14 assertions)
  ✓ structure.sql (0 assertions) <--- this is the file that causes segfault in neovim

```

matthias-Q:
fixes #221 

I think I have found the issue, but I suspect that there are more since we are missing a lot of bounds checks in the `scanner.c`.

We should also avoid `strcpy` in favour of `strncpy` or even better `memcpy`. But I am by no means an expert in C.

Steps to reproduce:
(I use `tree-sitter-cli` from crates.io)
1. Setup the tree-sitter-cli config in `~/.config/tree-sitter/config.json` and specify the parser-directories. This should point to this repo.
2. run `tree-sitter generate` to create the `parser.c`
3. compile with `cc -shared -O -D_FORTIFY_SOURCE=3 -fPIE -fPIC -I./src src/parser.c src/scanner.c -o sql.so`
4. run `valgrind tree-sitter parse <statements.sql>`

The examples from issue #221 reveal that in line 185 (almost at the end) there was a `malloc` with `length - 1`

Valgrind reports some more issues though. 

On my second machine the issue was not reproducable. I suspect the versions of gcc and glibc. On this machine I have `GNU C Library (GNU libc) stable release version 2.38.` and `cc --version
cc (GCC) 13.2.1 20230801`

On my second machine, I have older versions (gcc 11, glibc 2.35) and it did not segfault.

one more note. The SQL statement in the issue is not parsed correctly anyway. It just does not segfault anymore.

### DIFF
--- a/src/scanner.c
+++ b/src/scanner.c
@@ -28,20 +28,20 @@ void *tree_sitter_sql_external_scanner_destroy(void *payload) {
     state->start_tag = NULL;
   }
   free(payload);
+  return NULL;
 }
 
-char* add_char(char* text, int* text_size, char c, int index) {
+char* add_char(char* text, size_t* text_size, char c, int index) {
   if (text == NULL) {
     text = malloc(sizeof(char) * MALLOC_STRING_SIZE);
     *text_size = MALLOC_STRING_SIZE;
-    text[index] = c;
-    text[index + 1] = '\0';
   }
 
-  if (index + 1 == *text_size) {
+  // will break when indexes advances more than MALLOC_STRING_SIZE
+  if (index + 1 >= *text_size) {
     *text_size += MALLOC_STRING_SIZE;
     char* tmp = malloc(*text_size * sizeof(char));
-    strcpy(tmp, text);
+    strncpy(tmp, text, *text_size);
     free(text);
     text = tmp;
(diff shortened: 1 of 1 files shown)


kirillrdy:
@matthias-Q I can confirm that this fixes segfault for me !

mortezadadgar:
yes this fixes the issue
I like to mention a few things:
1. they're some useless `malloc` in the code that can be removed like (`text_size`) but this is not related to this pr
2. other `strcpy` usages can be replaced with `strncpy` as well

DerekStride (approved):
Awesome work!

PR #227, opened 2023-11-22
feat: Add REFERENCES column constraint to CREATE TABLE statements

dgmcdona:
See [postgresql CREATE TABLE docs](https://www.postgresql.org/docs/current/sql-createtable.html)
```
where column_constraint is:

[ CONSTRAINT constraint_name ]
{ NOT NULL |
  NULL |
  CHECK ( expression ) [ NO INHERIT ] |
  DEFAULT default_expr |
  GENERATED ALWAYS AS ( generation_expr ) STORED |
  GENERATED { ALWAYS | BY DEFAULT } AS IDENTITY [ ( sequence_options ) ] |
  UNIQUE [ NULLS [ NOT ] DISTINCT ] index_parameters |
  PRIMARY KEY index_parameters |
  REFERENCES reftable [ ( refcolumn ) ] [ MATCH FULL | MATCH PARTIAL | MATCH SIMPLE ]
    [ ON DELETE referential_action ] [ ON UPDATE referential_action ] }
[ DEFERRABLE | NOT DEFERRABLE ] [ INITIALLY DEFERRED | INITIALLY IMMEDIATE ]
```

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -2487,6 +2487,27 @@ module.exports = grammar({
         $.keyword_null,
         $._not_null,
       ),
+      seq(
+        $.keyword_references,
+        $.object_reference,
+        paren_list($.identifier, true),
+        repeat(
+          seq(
+            $.keyword_on,
+            choice($.keyword_delete, $.keyword_update),
+            choice(
+              seq($.keyword_no, $.keyword_action),
+              $.keyword_restrict,
+              $.keyword_cascade,
+              seq(
+                $.keyword_set,
+                choice($.keyword_null, $.keyword_default),
+                  optional(paren_list($.identifier, true))
+              ),
+            ),
+          ),
+        ),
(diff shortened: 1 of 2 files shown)


matthias-Q (approved):
LGTM

matthias-Q:
@DerekStride not sure if I did everything correctly on the release. I have create a release branch which has the updated changelog.

DerekStride:
@matthias-Q hmm, I think I pushed directly to main when I made the first release. Instead we might want to update the script to not create the git tag. That way we can have GitHub create it during the creation of the release.

matthias-Q:
Hmm, ok. I cannot push to main (which is a good idea). Should we merge the new branch to main in order to keep the changelog & package.json in sync?

DerekStride:
Yeah, let's do that

PR #229, opened 2023-11-24
feat!: Add complex index fields to CREATE INDEX statements

dgmcdona:
See [postgresql CREATE INDEX docs](https://www.postgresql.org/docs/current/sql-createindex.html)

The CREATE INDEX statement should be able to support three different types of index fields:
1. Column names
2. Expressions (wrapped in parentheses)
3. Function calls

This PR adds those features to the CREATE INDEX statement while preserving backwards compatibility with existing tests/queries. 

There is some code duplication going on here that I couldn't figure out how to resolve; maybe someone has a suggestion for how this can be improved. Each of the three types of index fields takes a set of optional parameters, but I couldn't break them out into their own rule since they are all optional and a rule containing only optional fields would match the empty string rule and therefore be invalid. I also couldn't wrap the three types in a `choice` followed by the optional nodes since the optional nodes need to be children of the `column` node in order to not break existing tests/queries.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1301,6 +1301,35 @@ module.exports = grammar({
       alias($._literal_string, $.literal),
     ),
 
+    _operator_class: $ => seq(
+      field("opclass", $.identifier),
+      optional(
+        field("opclass_parameters", wrapped_in_parenthesis(comma_list($.term)))
+      )
+    ),
+
+    _index_field: $ => seq(
+      choice(
+        field("expression", wrapped_in_parenthesis($._expression)),
+        field("function", $.invocation),
+        field("column", $._column),
+      ),
+      optional(seq($.keyword_collate, $.identifier)),
+      optional($._operator_class),
+      optional($.direction),
+      optional(
+        seq(
+          $.keyword_nulls,
+          choice(
+            $.keyword_first,
+            $.keyword_last
+          )
+        )
+      ),
+    ),
+
+    index_fields: $ => wrapped_in_parenthesis(comma_list(alias($._index_field, $.field))),
+
     create_index: $ => seq(
       $.keyword_create,
       optional($.keyword_unique),
@@ -1329,7 +1358,7 @@ module.exports = grammar({
             ),
           ),
         ),
-        $.ordered_columns,
+        $.index_fields
       ),
       optional(
         $.where,
--- a/test/corpus/create.txt
+++ b/test/corpus/create.txt
@@ -1218,9 +1218,9 @@ CREATE UNIQUE INDEX "akRoleName" ON "Role" ("name");
       (keyword_on)
       (object_reference
         name: (identifier))
-      (ordered_columns
-        (column
-          name: (literal))))))
+      (index_fields
+        (field
+          column: (literal))))))
 
 ================================================================================
 Create table as select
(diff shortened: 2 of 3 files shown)


dgmcdona:
This works as well and is a little more concise, but changes the output slightly for the test that I added: 
```javascript
    _index_field: $ => seq(
      choice(
        field("column_expression", wrapped_in_parenthesis($._expression)),
        field("column_function", $.invocation),
        field("name", $._column),
      ),
      optional(seq($.keyword_collate, $.identifier)),
      optional($._operator_class),
      optional($.direction),
      optional(
        seq(
          $.keyword_nulls,
          choice(
            $.keyword_first,
            $.keyword_last
          )
        )
      ),
    ),

    index_fields: $ => wrapped_in_parenthesis(comma_list(alias($._index_field, $.column))),

    create_index: $ => seq(
      $.keyword_create,
      optional($.keyword_unique),
      $.keyword_index,
      optional($.keyword_concurrently),
      optional(
        seq(
          optional($._if_not_exists),
          field("column", $._column),
        ),
      ),
      $.keyword_on,
      optional($.keyword_only),
      seq(
        $.object_reference,
        optional(
          seq(
            $.keyword_using,
            choice(
              $.keyword_btree,
              $.keyword_hash,
              $.keyword_gist,
              $.keyword_spgist,
              $.keyword_gin,
              $.keyword_brin
            ),
          ),
        ),
        alias($.index_fields, $.ordered_columns)
      ),
      optional(
        $.where,
      ),
    ),
```

grammar.js
```diff
@@ -1329,7 +1394,7 @@ module.exports = grammar({
             ),
           ),
         ),
-        $.ordered_columns,
+        alias($.index_fields, $.ordered_columns)
```
DerekStride, line 1397: I think we can just change this to be `index_fields`

```suggestion
        $.index_fields
```

grammar.js
```diff
@@ -1301,6 +1301,71 @@ module.exports = grammar({
...
+
+    _index_field: $ => seq(
+      choice(
+        alias($._index_column, $.column),
+        alias($._index_expression, $.column_expression),
+        alias($._index_column_function, $.column_function)
+      )
+    ),
+
+    index_fields: $ => wrapped_in_parenthesis(comma_list($._index_field)),
```
DerekStride, line 1367:
```suggestion
    index_fields: $ => wrapped_in_parenthesis(comma_list(alias($._index_field, $.field))),
```

grammar.js
```diff
@@ -1301,6 +1301,71 @@ module.exports = grammar({
...
+          )
+        )
+      )
+    ),
+
+    _index_field: $ => seq(
+      choice(
+        alias($._index_column, $.column),
+        alias($._index_expression, $.column_expression),
+        alias($._index_column_function, $.column_function)
```
DerekStride, line 1363: I like your suggestion of moving the duplication into this `_index_field` node. I'd suggest using a `field` for each type of field like this:

```suggestion
        field("expression", wrapped_in_parenthesis($._expression)),
        field("function", $.invocation),
        field("column", $._column),
```

DerekStride (commented):
I like your suggested change but I'd also make a few more. I don't think we need to worry about breaking changes to the AST being disruptive especially for the `create_index` node. We should still follow the conventional commit guidelines here for breaking changes though.

cc: @dmfay @matthias-Q

dgmcdona:
Thanks for the feedback! I made those changes and updated the commit message/PR title to reflect breaking changes.

DerekStride:
I merged this to move things forward but if there's any concerns I can follow up in another PR

PR #235, opened 2024-01-01
fix(scanner): do not allow whitespaces in dollar quote tag

eklmv:
Whitespace characters not allowed in dollar quoted tags.
[docs](https://www.postgresql.org/docs/16/sql-syntax-lexical.html#SQL-SYNTAX-DOLLAR-QUOTING):
> The tag, if any, of a dollar-quoted string follows the same rules as an unquoted identifier, except that it cannot contain a dollar sign.

Fixes problems with [sqlc](https://github.com/sqlc-dev/sqlc) queries, which use positional parameters outside function definitions or prepared statements. For example:
```sql
SELECT * FROM a
LIMIT $1 OFFSET $2;

SELECT * FROM b
LIMIT $1 OFFSET $2;
```
`$1 OFFSET $` should not be recognized as a tag name.

### DIFF
--- a/test/corpus/errors.txt
+++ b/test/corpus/errors.txt
@@ -149,37 +149,43 @@ Function body with a tag name contains whitespace
 create or replace function public.do_stuff()
   returns trigger
   language plpgsql
-as $a $
+as $a$
 begin
-  return $b$text$b$;
+  return $b $text$b $;
 end;
-$a $;
+$a$;
 
 --------------------------------------------------------------------------------
 
 (program
-  (ERROR
-    (keyword_create)
-    (keyword_or)
-    (keyword_replace)
-    (keyword_function)
-    (object_reference
-      (identifier)
-      (identifier))
-    (function_arguments)
-    (keyword_returns)
-    (keyword_trigger)
-    (function_language
-      (keyword_language)
-      (identifier))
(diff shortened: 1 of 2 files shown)


DerekStride (approved):

matthias-Q (approved):

test/corpus/errors.txt
```diff
@@ -141,3 +141,45 @@ $c$;
...
+    (function_arguments)
+    (keyword_returns)
+    (keyword_trigger)
+    (function_language
+      (keyword_language)
+      (identifier))
+    (ERROR
+      (keyword_as)
+      (UNEXPECTED 'a')
+      (UNEXPECTED '\n'))
```
matthias-Q, line 178: well, this will not work on windows machines. 
any ideas @DerekStride ?

DerekStride: Could the test be written as `as $a $ begin` ?

eklmv: I tried `as $a $ begin` but it produced strange behavior with `(UNEXPECTED ')`, after `test --update` test just fails with diff which doesn't contains expected/actual.
Changing `$b $text$b $; ` doesn't capture any newlines and works.

PR #248, opened 2024-04-12
add: `check constraint` for column definitions

### LINKED ISSUE #247: Support for check constraints
The parser does not currently seem to support check constraints, for example:

```sql
CREATE TABLE products (
    product_no integer,
    name text,
    price numeric CHECK (price > 0)
);
```

For reference: https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-CHECK-CONSTRAINTS

matthias-Q:
fixes #247 

@DerekStride I think I have a newer version of tree-sitter on my machine and I get a bunch of warnings. There was a change about the order of query matches (which breaks the tests) and something related to non static functions in the scanner.c

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -2544,6 +2544,7 @@ module.exports = grammar({
       $.keyword_auto_increment,
       $.direction,
       $._column_comment,
+      $._check_constraint,
       seq(
         optional(seq($.keyword_generated, $.keyword_always)),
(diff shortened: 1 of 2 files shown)


DerekStride (approved):
I'll take a look at the changes from the latest cli and we can try to upgrade to the latest

Issue #234, opened 2023-12-15
Include parser files in Release Assets


DerekStride:
# What

Currently the only assets included on the release are the source files (see https://github.com/DerekStride/tree-sitter-sql/releases/tag/v0.2.0). We should generate & add the parser files to the release as well. We could even precompile the parser for difference architectures and make that available too.

- [ ] Add workflow to generate & add parsers files to release assets
- [ ] Add workflow to precompile the parser & add to release assets
  - [ ] x86_64-darwin-20
  - [ ] x86_64-linux

qoops-1:
It would also be great to have them in the repo itself, not just the assets. Emacs has a built-in workflow to install tree-sitter grammars, which clones the repo at a specific tag and looks for `parser.c` file there. Because the file is not present here, it doesn't work with this repo

matthias-Q:
@qoops-1 have a look at #120 We have discussed this issue already with the emacs guys. The problem is, that the `parser.c` is auto generated and gets quite large.

qoops-1:
Thank you for the link, my takeaway from it is that current approach is to track hashes on `gh-pages` for corresponding tags of releases on `main`. It would be easier if `gh-pages` was also tagged

clason:
On a related note, could you include the `grammar.js` in the `gh-pages` branch? Omitting this breaks nvim-treesitter's `TSInstallFromGrammar`.

PR #253, opened 2024-04-21
feat(ci): add workflow to include parser and binary artifacts

matthias-Q:
This PR adds a GH Workflow to build source artifacts including the  `parser.c`, `sql.so` and a wasm file for Linux x86_64 and MacOS Arm64.

Along the way, I have fixed an issue in the `parser.c` (replaced `strcpy` with `memcpy`)

As of now, this does probably not work as intended, as in the final step the github cli tool is used to create a release and the trigger is the manual creation of the release. I have no idea how to fix that. I fails due to a collision of release tags. @DerekStride do you have any ideas? 

This PR is aimed to solve #234

I have noticed some issues: When installing it via npm, we require the `nan` package. It was n the lock file, but to get it working I had to add a manual installation step beforehand.


matthias-Q:
I did some changes in the mean time and I think this could work.

We keep the workflow as outlined in the CONTRIBUTION.md:
* create a release PR (after running commit-tag-version)*
* the new workflow `create_release.yml` will only trigger when a tagged release on `main` has been made the follows the semver pattern. This removed the need for manually creating a release.

I tried to test that on my fork:

https://github.com/matthias-Q/tree-sitter-sql/actions/runs/8773987845

clason:
For the record, upstream is planning to add support for that to the CLI (`tree-sitter publish`); see also https://github.com/tree-sitter/workflows.

(And isn't the wasm file platform independent? I thought that was the whole point!)

matthias-Q:
Yeah, I think wasm is platform independent. I just included it in the build process for each system. So maybe I should add it separately. 

I will take a look into `tree-sitter publish`. Thanks for the hint.

matthias-Q:
I have separated the wasm build and cleaned up the pipeline a bit.

DerekStride:
> @DerekStride do you have any ideas?

I think we have to create a draft release and upload artifacts to that on push to main.

Then it'd be a manual process to convert the draft to an actual release.

We'd end up with a bunch of draft releases which is probably fine but we could have a cleanup workflow.

matthias-Q:
IIUC in my latest version, the release pipeline will only be triggered once a new tagged commit on main is made. It just makes the manual release MR unnecessary. But please double check that, as it is quite hard to test. 
I am referring to this CI trigger.
´´´
on:
  push:
    branch:
      - main
    tags:
      - v[0-9]+.[0-9]+.[0-9]+
´´´

PR #257, opened 2024-06-03
add: `copy from stdin`

### LINKED ISSUE #256: Support for COPY FROM STDIN
For example:

```sql
COPY some_table (a, b, c) FROM STDIN (FORMAT CSV);
1,2,3
d,e,f
\.
```

The actual data used in my situation is massive, and it causes the parsing to take a very long time (over a minute to open the file in neovim). I assume it would be much more efficient if the parser didn't attempt to parse the CSV data as SQL. 

I believe the ideal behavior would be to treat the data in between the `;` and `\.` as `csv` if the format is explicitly CSV, and text otherwise. Parsing other components of the `COPY` statement would be nice, but for me personally, handling the data contents would be the biggest win. 

Ref: https://www.postgresql.org/docs/current/sql-copy.html

matthias-Q:
adds `COPY FROM STDIN` parsing

Closes #256


DerekStride (approved):
Let's merge this and tackle the data part later. I'll leave some questions on the issue

PR #260, opened 2024-06-24
Fix/update bindings

grpse:
# Summary

trying to use inside of a node project, the binding when calling `setLanguage` from node tree-sitter, was throwing an error `Type error: Invalid language object`. This was due to the use of `nan` in the old binding from tree-sitter, but now it uses node-addon-api, so I've updated it. Also, for wasm build, the scanner.c uses strcpy, but it is not available symbol after compilation with new version of tree-sitter.


package.json
```diff
@@ -1,22 +1,34 @@
...
-    "nan": "^2.18.0"
+    "node-addon-api": "^7.1.0",
+    "node-gyp-build": "^4.8.0"
   },
   "devDependencies": {
     "commit-and-tag-version": "^12.0.0",
     "node-gyp": "^10.0.1",
-    "tree-sitter-cli": "^0.20.8"
+    "prebuildify": "^6.0.0",
+    "tree-sitter-cli": "^0.22.6"
```
matthias-Q, line 23: Good that we upgrade. However it will break the tests for now, since the queries in the tests are interpreted in a different order (I don't remember if first or last query wins)  regardless, we have to make changes in the tests.

grpse: Yes, tested and it inversed the type and function.call. I'm fixing it.

Makefile
```diff
@@ -1,4 +1,4 @@
-CC ?= gcc
+CC ?= clang++
```
DerekStride, line 1: can we keep the default to `gcc` I think that's more common

binding.gyp
```diff
@@ -1,19 +1,31 @@
...
+                "<!(node -p \"require('node-addon-api').targets\"):node_addon_api_except",
+            ],
+            "include_dirs": [
+                "src",
+            ],
+            "sources": [
+                "bindings/node/binding.cc",
+                "src/parser.c",
+                "src/scanner.c",
+                # NOTE: if your language has an external scanner, add it here.
```
DerekStride, line 15:
```suggestion
```

alexbit-codemod:
thanks for working on this.... once this lands, we can integrate sql parser to ast-grep to enable automated code transformation of sql for a variety of use cases. so looking forward to it.

Makefile
grpse: done

binding.gyp
grpse: done

package.json
grpse: fixed. The highlight evaluation is switched, so I changed it and now tests are passing.

DerekStride (approved):

Issue #269, opened 2024-08-05
Publish on PyPI?


prurph:
This grammar isn't available for installation in Python projects via `pip`. The tree-sitter team has a workflow [here](https://github.com/tree-sitter/workflows?tab=readme-ov-file#pypi-options); what do you think about adding this to the project to publish to PyPi? Thanks for considering!

DerekStride:
Done! It's available here: https://pypi.org/project/tree-sitter-sql/0.3.5/

prurph:
Wow what a fast response! Thank you so much!

matthias-Q:
I just tried it out and 
```python
tree_sitter_sql.language()
```
returns an integer for me and not a `Language`

not sure if that is correct

PR #270, opened 2024-08-06
feat: publish package to pypi

DerekStride:
## What

Publish package to pypi, part of #269. I'm testing out the [tree-sitter/workflows](https://github.com/tree-sitter/workflows), if it works without the generated parser files being present we can try migrating the other workflows.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -32,3 +32,7 @@ jobs:
       - uses: katyo/publish-crates@v2
         with:
           registry-token: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+  pypi-publish:
+    uses: tree-sitter/workflows/.github/workflows/package-pypi.yml@main
+    secrets:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -32,3 +32,7 @@ jobs:
       - uses: katyo/publish-crates@v2
         with:
           registry-token: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+  pypi-publish:
+    uses: tree-sitter/workflows/.github/workflows/package-pypi.yml@main
```
matthias-Q, line 36: do we trust `@main` or do you want to pin it to a hash?

matthias-Q (approved):

.github/workflows/publish.yml
DerekStride: I think main is fine so that we benefit from any fixes. If it causes trouble we can always pin it later

Issue #282, opened 2024-10-23
Syntax Highlighting is broken for tree-sitter 0.24


jakob:
When I use the command `tree-sitter highlight test.sql` with this grammar, I get the following error:
```
Error in query file "queries/highlights.scm"

Caused by:
    Query error at 151:4. Invalid node type keyword_min
```

(It works with tree-sitter 0.23.2, but it is broken in 0.24 or newer)

I believe it is caused by a new feature where `tree-sitter generate` removes unused rules. Since `keyword_max` is not used in the grammar, it is removed, and the highlight query fails because it uses an unknown node type.

DerekStride:
closed by #285

PR #283, opened 2024-10-30
tree-sitter 0.24 upgrade

mwhitworth:
needs #279 

changes required to upgrade to tree-sitter 0.24. But it looks like there is at least one issue with a 0.24 upgrade (#282), and upgrading to 0.23 has the advantage of moving Rust bindings to `tree-sitter-language` and simplifying further upgrades there.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,20 +1,18 @@
 [package]
 name = "tree-sitter-sequel"
-description = "sql grammar for the tree-sitter parsing library"
+description = "Tree-sitter Grammar for SQL"
 version = "0.3.5"
-keywords = ["incremental", "parsing", "sql"]
-categories = ["parsing", "text-editors"]
-repository = "https://github.com/DerekStride/tree-sitter-sql"
-edition = "2018"
+authors = ["derek stride"]
 license = "MIT"
+readme = "README.md"
+keywords = ["incremental", "parsing", "tree-sitter", "sql"]
+categories = ["parsing", "text-editors"]
+repository = "git+https://github.com/derekstride/tree-sitter-sql.git"
+edition = "2021"
+autoexamples = false
 
 build = "bindings/rust/build.rs"
-include = [
-  "bindings/rust/*",
-  "grammar.js",
-  "queries/*",
-  "src/*",
-]
+include = ["bindings/rust/*", "grammar.js", "queries/*", "src/*"]
 
 [lib]
 path = "bindings/rust/lib.rs"
@@ -23,7 +21,7 @@ path = "bindings/rust/lib.rs"
 tree-sitter-language = "0.1"
 
 [build-dependencies]
-cc = "~1.2.1"
+cc = "1.1.22"
 
 [dev-dependencies]
-tree-sitter = "0.23"
+tree-sitter = "0.24.3"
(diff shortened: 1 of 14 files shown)


matthias-Q:
@mwhitworth can you rebase `main` into your branch please? I have updated CI and also bumped tree-sitter to 0.23. CI should pass with tree-sitter 0.24 as well.

Once this lands I would make a new release.

mwhitworth:
I'm rebasing this now (and checking the diff)

I came across a message while cloning the repository, and some odd behaviour around `bindings/swift/TreeSitterSQL/sql.h`. This is because there are actually two copies of the file in the working tree. Running `git clone` on the repo displays the following message:

```
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree
  'bindings/swift/TreeSitterSQL/sql.h'
  'bindings/swift/TreeSitterSql/sql.h'
```

It might be clearest to fix this separately.

Cargo.toml
```diff
@@ -23,7 +21,7 @@ path = "bindings/rust/lib.rs"
 tree-sitter-language = "0.1"
 
 [build-dependencies]
-cc = "~1.2.1"
```
matthias-Q, line 26: I think we should go with ~1.2.1

bindings/rust/build.rs
```diff
@@ -2,23 +2,21 @@ fn main() {
...
     #[cfg(target_env = "msvc")]
     c_config.flag("-utf-8");
 
     let parser_path = src_dir.join("parser.c");
     c_config.file(&parser_path);
+    println!("cargo:rerun-if-changed={}", parser_path.to_str().unwrap());
 
-    // If your language uses an external scanner written in C,
-    // then include this block of code:
+    // NOTE: if your language uses an external scanner, uncomment this block:
```
matthias-Q, line 14: I think we should do this. But maybe it would be also a separate PR

matthias-Q (changes requested):
LGTM. Can you make these two changes?

mwhitworth:
> ....
> It might be clearest to fix this separately.

https://github.com/DerekStride/tree-sitter-sql/pull/290

Then I'll finalize this branch for merge.

bindings/rust/build.rs
mwhitworth: I'll fix this in this branch.

(It also makes me curious about making the upgrades easier in `tree-sitter` itself, given that e.g. filesystem state should be easy to inspect during upgrades)

Package.swift
```diff
@@ -1,34 +1,37 @@
 // swift-tools-version:5.3
-
 import PackageDescription
 
 let package = Package(
-    name: "TreeSitterSQL",
-    platforms: [.macOS(.v10_13), .iOS(.v11)],
+    name: "TreeSitterSql",
```
mwhitworth, line 5: Two meaningful changes to note:
- the Swift package name has been changed - I don't know if capitalization matters for this package manager
- the `exclude` paths have disappeared, but the effect on the package should be minimal.

mwhitworth:
With this rebase and some further `fix` commits, this should be good to go. Ideally there should be less `fix` commits for each `tree-sitter` release, but I suspect this is mostly an upstream issue. Since I've helped to apply `tree-sitter` upgrades across some other language grammars, I'll be paying attention to this.

matthias-Q (approved):

matthias-Q:
Thanks @mwhitworth for the effort!

PR #285, opened 2024-11-14
bump: action versions

matthias-Q:
Bumping action versions from 3 to 4. Maybe that solves the issue that tests are failing on windows machines in CI


DerekStride (approved):

matthias-Q:
@DerekStride bumping it to v4 did not do the trick. For some reason, with the new tree-sitter version, on windows npm tries to find the tests at
`'D:\a\tree-sitter-sql\tree-sitter-sql\bindings\node\*_test.js'` 
While on unix machines it takes a relative path. I suspect that the double "tree-sitter-sql" directories are doing this. I have no clue where that comes from. Mainly because I rarely use npm :-)

matthias-Q:
Looks better now. Tests are failing because of `highlights.scm` `keyword_max` seems to be an invalid node type. No clue why.

matthias-Q:
@DerekStride can you take a second look? This MR may supersede #279 

I have disabled tests for swift and go as I could not get them running. Neither on my machine nor in CI.

DerekStride (approved):
LGTM, thanks for migrating to the new `tree-sitter/` actions

DerekStride:
Not sure why it's waiting for `windows-2019` might just be a bug, I think we can ignore and merge without it

matthias-Q:
@DerekStride I am using windows-2019 again instead of windows-latest. I think a merge rule was preventing this. If we want to use windows-latest, then you probably have to check the Merge rules again (and rename it from windows-2019 to windows-latest).

matthias-Q:
@DerekStride any idea why jekyll fails now?

PR #288, opened 2024-11-15
feat: add `cross join unnest with ordinality`

### LINKED ISSUE #284: [Presto] Parser error when it encounters WITH ORDINALITY statement of UNNEST
WITH ORDINALITY is used by the [UNNEST](https://prestodb.io/docs/current/sql/select.html#unnest) function in Presto.
The following code snippet fails:

```
SELECT numbers, n, a
FROM (
  VALUES
    (ARRAY[2, 5]),
    (ARRAY[7, 8, 9])
) AS x (numbers)
CROSS JOIN UNNEST(numbers) WITH ORDINALITY AS t (n, a);
```

matthias-Q:
Closes #284 

I am not so sure about the alias at the end. 
`relation` can already have an alias, but it does not support this unpacking of arrays. So this compiles, but the resulting AST might be misleading-

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -294,6 +294,7 @@ module.exports = grammar({
     keyword_out: _ => make_keyword("out"),
     keyword_inout: _ => make_keyword("inout"),
     keyword_variadic: _ => make_keyword("variadic"),
+    keyword_ordinality: _ => make_keyword("ordinality"),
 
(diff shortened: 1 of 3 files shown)


DerekStride (approved):
LGTM, we can always improve it later if it's not sufficient

PR #293, opened 2024-11-19
fix: bump other versions to 0.3.6 and add to docs

DerekStride:
## What

I forgot that we need to manually update the version numbers for some of the other dependencies. I added a mention of it to the CONTRIBUTING.md docs so we don't forget again.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.13)
 
 project(tree-sitter-sql
-        VERSION "0.3.5"
+        VERSION "0.3.6"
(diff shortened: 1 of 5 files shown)


DerekStride:
@matthias-Q I updated the version numbers here but there was a bug in the pypi release workflow and I put up a PR to fix it here: https://github.com/tree-sitter/workflows/pull/19

After that's merged we can retry the release workflow.

mwhitworth:
I see the `tree-sitter-sequel` Rust crate version still at v0.3.5 - ref https://crates.io/crates/tree-sitter-sequel/

matthias-Q:
@DerekStride do you want to do a follow up to trigger the release of python/rust packages to >0.3.5?

PR #301, opened 2025-02-01
Move tree-sitter-cli to regular dependencies to avoid installation error

jens-f:
Move tree-sitter-cli to regular dependencies to avoid installation error

My package json looks like this:
```
...
  "dependencies": {
    "@derekstride/tree-sitter-sql": "^0.3.7",
    "tree-sitter": "^0.21.0"
  }
...
```
When running `npm i` the installation fails because `tree-sitter-sql` is attempting to run this command `"tree-sitter generate && node-gyp-build"`. The `tree-sitter` cli command is not available. In order to avoid this, `tree-sitter-sql` needs to be a regular dependency instead of a `devDependency`

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,13 +16,13 @@
   "license": "MIT",
   "dependencies": {
     "node-addon-api": "^7.1.0",
-    "node-gyp-build": "^4.8.0"
+    "node-gyp-build": "^4.8.0",
+    "tree-sitter-cli": "^0.24.0"
   },
   "devDependencies": {
     "commit-and-tag-version": "^12.0.0",
     "node-gyp": "^10.0.1",
-    "prebuildify": "^6.0.0",
-    "tree-sitter-cli": "^0.24.0"
+    "prebuildify": "^6.0.0"
   },
   "peerDependencies": {
     "tree-sitter": "^0.21.0"


DerekStride (commented):
`tree-sitter-cli` is a build dependency but shouldn't be required at runtime. But this would make using it easier, I'm on the fence about this change. Thoughts @matthias-Q @dmfay?

dmfay:
I think instead we can use `npx tree-sitter` in the package.json scripts and npm will use the binary installed through devDependencies, but I haven't tested this.

jens-f:
As I understand it, `npx` would try to execute the `tree-sitter` command from the `devDependencies`, _if_ they were present. But given that `npm i` doesn't seem to install the `devDependencies`, it will currently fail. But you can still make it work by providing the package name to npx, which then would install the cli package in the npm cache directory if not yet present. 
```
npx --yes --package=tree-sitter-cli -- tree-sitter generate && node-gyp-build
```
The end-result is that the tree-sitter dependency is still installed during the `npm i` run, just in the cache instead of the local dependencies as proposed by this PR.

Would you like me to change the PR to that version?

DerekStride:
If the `--package` option would remove the need for the `npm link` in #302 I think that would probably be the best option cc @dmfay

dmfay:
> As I understand it, `npx` would try to execute the `tree-sitter` command from the `devDependencies`, _if_ they were present. But given that `npm i` doesn't seem to install the `devDependencies`, it will currently fail. But you can still make it work by providing the package name to npx, which then would install the cli package in the npm cache directory if not yet present.
> 
> ```
> npx --yes --package=tree-sitter-cli -- tree-sitter generate && node-gyp-build
> ```
> 
> The end-result is that the tree-sitter dependency is still installed during the `npm i` run, just in the cache instead of the local dependencies as proposed by this PR.
> 
> Would you like me to change the PR to that version?

yeah, I think I prefer that -- keeping the cli out of node_modules sounds like the best available option.

jens-f:
I updated the PR with the npx use for all scripts. I also removed the `devDependency` for the cli since it doesn't seem to be used anymore. But please let me know if you want to keep that dependency

jens-f:
Some of the tests are failing but as far as I can tell they don't seem to be related to this PR.

DerekStride:
I went ahead and merged, I can tackle fixing the cargo tests in a separate PR

jens-f:
Thanks @DerekStride  and @dmfay for accepting the PR and merging it. Would it be possible to make a new npm release of `tree-sitter-sql` that includes this change?

DerekStride:
`v0.3.8` is now published with the fixes: https://www.npmjs.com/package/@derekstride/tree-sitter-sql

PR #309, opened 2025-03-07
fix: module path in binding_test.go

rob2244:
Fix the module path in binding_test.go to fix issues that occur when installing the bindings from golang this addresses #308


DerekStride (approved):

DerekStride:
I'll try to setup a working CI suite for the go bindings when I have time

Issue #310, opened 2025-03-11
Missing optional database name in object_reference rule


rob2244:
The object reference rule should have an optional database specifier before the schema, for example for this sql 

`SELECT * FROM blizz.bazz.foo`

I expect to get the following syntax tree:
(program (statement (select (keyword_select) (select_expression (term value: (all_fields)))) (from (keyword_from) (relation (object_reference database: (identifier) schema: (identifier) name: (identifier))))))

But get the following instead:
(program (statement (keyword_with) (cte (identifier) (keyword_as) (statement (select (keyword_select) (select_expression (term value: (all_fields)))) (from (keyword_from) (relation (ob
ject_reference schema: (identifier) name: (identifier)) (ERROR) alias: (identifier)) (where (keyword_where) predicate: (binary_expression left: (field name: (identifier)) right: (liter
al)))))) (select (keyword_select) (select_expression (term value: (all_fields)))) (from (keyword_from) (relation (object_reference schema: (identifier) name: (identifier)) (ERROR) alia
s: (identifier)))))

rob2244:
I'll create a pr for this in a little bit

rob2244:
Closing this issue as the relevant pr has been merged

PR #311, opened 2025-03-11
fix(ast): Add database name to object_reference

rob2244:
Add optional database name to the object_reference rule and add some tests to add the new functionality

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -2195,12 +2195,18 @@ module.exports = grammar({
       $.identifier,
     ),
 
-    object_reference: $ => seq(
-      optional(
-        seq(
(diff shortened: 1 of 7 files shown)


rob2244:
Resolves #310

rob2244:
Realized I didn't take care of some of the steps in the contributing.md like the version bump. Sorry will do tomorrow

DerekStride:
We do adhoc releases at the moment so the version bump isn't necessary unless you'd like us to publish a new version to npm, crates, or pypi. That can be in a follow-up PR if you like – this one looks great, thank you 🎉

Issue #236, opened 2024-01-14
Syntax highlighting appears to give up in Postgres function bodies


jaidetree:
Given the following Postgres SQL function:

```sql
CREATE OR REPLACE FUNCTION start_task(p_task_name VARCHAR)
RETURNS sessions AS $$
DECLARE
  task_id INTEGER;
  task_row RECORD;
  session_record sessions;
BEGIN
    -- Look for any ongoing sessions for an active task
    SELECT *
      INTO task_row
    FROM tasks t
    INNER JOIN sessions s ON t.id = s.task_id
    WHERE s.end_time IS NULL
    AND s.is_active = TRUE
    AND t.is_active = TRUE
    LIMIT 1;

    -- If an ongoing session was found, throw an error with hint
    IF task_row.id IS NOT NULL THEN
      RAISE EXCEPTION 'Task % "%" is still in progress. Action refused.',
        task_row.id, task_row.task_name
    ELSE
      -- Look up the task_id from the name
      SELECT id INTO task_id
      FROM tasks
      WHERE task_name = p_task_name;

      -- Start the session
      INSERT INTO sessions (task_id)
      VALUES (task_id)
      RETURNING * INTO session_record;
    END IF;

    RETURN session_record;
END;
$$ LANGUAGE plpgsql;
```

Syntax highlighting completely stops for the rest of the file if I keep that semicolon after the first query that ends with `LIMIT 1;`

<img width="968" alt="image" src="https://github.com/DerekStride/tree-sitter-sql/assets/590297/a3a90ceb-a81a-4595-945f-a711ced4f535">

If I remove that semicolon after the first query the syntax highlighter continues highlighting the file but the syntax is invalid and running the migration fails.

Any ideas what's causing that? Happy to help with a PR to fix but will need some guidance. Though as far as I know, I'm just using this wrong 😅

dmfay:
[...]

PR #313, opened 2025-04-08
Support for T-SQL

stackmystack:
Hello Derek,

I would like to add support for Microsoft's T-SQL if possible.

This is a sketch of what it would look like. 

I am not happy with the PR's state at all, but I thought I'd share what I have as of now, and collect feedback. I will re-write once, and if, support for T-SQL is accepted, and we agree on how to organise things; I am not proficient with all of the dialects of SQL so I may be implementing things in a stupid way.

The most important thing I would like to see, is recognising `@variable_name` from T-SQL as an `identifier` instead of having it in `tsql_parameter` as I do right now.

This PR is motivated by a real-world use case, which I can include in the test suite once I figure out how to make it work :) (node-gyp on my mac is not happy):

```sql
IF OBJECT_ID (N'dbo.LEVENSHTEIN_DISTANCE', N'FN') IS NOT NULL
    DROP FUNCTION dbo.LEVENSHTEIN_DISTANCE;
GO
CREATE FUNCTION dbo.LEVENSHTEIN_DISTANCE(@s nvarchar(4000), @t nvarchar(4000))
RETURNS int
AS
BEGIN
  DECLARE @sl int, @tl int, @i int, @j int, @sc nchar, @c int, @c1 int,
    @cv0 nvarchar(4000), @cv1 nvarchar(4000), @cmin int
  SELECT @sl = LEN(@s), @tl = LEN(@t), @cv1 = '', @j = 1, @i = 1, @c = 0
  WHILE @j <= @tl
    SELECT @cv1 = @cv1 + NCHAR(@j), @j = @j + 1
  WHILE @i <= @sl
  BEGIN
    SELECT @sc = SUBSTRING(@s, @i, 1), @c1 = @i, @c = @i, @cv0 = '', @j = 1, @cmin = 4000
    WHILE @j <= @tl
    BEGIN
      SET @c = @c + 1
      SET @c1 = @c1 - CASE WHEN @sc = SUBSTRING(@t, @j, 1) THEN 1 ELSE 0 END
      IF @c > @c1 SET @c = @c1
      SET @c1 = UNICODE(SUBSTRING(@cv1, @j, 1)) + 1
      IF @c > @c1 SET @c = @c1
      IF @c < @cmin SET @cmin = @c
      SELECT @cv0 = @cv0 + NCHAR(@c), @j = @j + 1
    END
    SELECT @cv1 = @cv0, @i = @i + 1
  END
  RETURN @c 
END
GO
```

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -55,6 +55,7 @@ module.exports = grammar({
             $.block,
           ),
           ';',
+          optional($.keyword_go),
         ),
       ),
       // optionally, a single statement without a terminating ;
@@ -266,6 +267,7 @@ module.exports = grammar({
     keyword_storage: _ => make_keyword("storage"),
     keyword_compression: _ => make_keyword("compression"),
     keyword_while: _ => make_keyword("while"),
+    keyword_go: _ => make_keyword("go"),
 
     keyword_trigger: _ => make_keyword('trigger'),
     keyword_function: _ => make_keyword("function"),


grammar.js
```diff
@@ -2675,11 +2790,11 @@ module.exports = grammar({
       ')',
     ),
 
-    column_definition: $ => seq(
+    column_definition: $ => prec.left(seq(
```
dmfay, line 2793: this is a _lot_ of "just prec.left it" and it's making me nervous

grammar.js
```diff
@@ -2177,6 +2263,14 @@ module.exports = grammar({
       optional(choice($.keyword_cascade, $.keyword_restrict)),
     ),
 
+    drop_function: $ => seq(
```
dmfay, line 2266: I'm pretty surprised we missed this, good catch!

grammar.js
```diff
@@ -2850,6 +2965,8 @@ module.exports = grammar({
       '*',
     ),
 
+
+    tsql_parameter: $ => /@([a-zA-Z_][0-9a-zA-Z_]*)/,
```
dmfay, line 2969: if my hazy memory of t-sql is anything to go by this might be more an identifier semantically, and making it an option for that would also fit a lot better with what you're doing in `function_argument` and `_qualified_field`

dmfay (changes requested):
please consider splitting this into two pull requests -- you've got a grab bag of fairly simple, straightforward improvements to t-sql support that are easy enough to evaluate on their own merits, but procedural sql and control flow is a [longstanding known gap](https://github.com/DerekStride/tree-sitter-sql/issues/236) that has much bigger ramifications for dialect support and performance.

also, needs more tests :)

stackmystack:
> please consider splitting this into two pull requests -- you've got a grab bag of fairly simple, straightforward improvements to t-sql support that are easy enough to evaluate on their own merits, 

Great, I will do so in a separate PR

> but procedural sql and control flow is a [longstanding known gap](https://github.com/DerekStride/tree-sitter-sql/issues/236) that has much bigger ramifications for dialect support and performance.

I am proposing these changes because I am pretty interested in procedural SQL and control flow. For the time being in T-SQL. I am putting the effort because I need them now. Would you be open to integrating them in `main` in the near future? In other words, should I fork the lib and not expect those to be accepted any time soon?

> 
> also, needs more tests :)

Definitely :)

grammar.js
stackmystack: Yes that's my impression too, I wanted confirmation from your part first.

vic-t: @stackmystack I'm curious, have you found a way to introduce bracketed identifiers?

PR #314, opened 2025-04-09
Mssql simple changes

stackmystack:
Following up on https://github.com/DerekStride/tree-sitter-sql/pull/313

This is the first batch of features.

Adds support for:

1. DROP FUNCTION
2. N'string'
3. @param
4. OBJECT_ID

I tried to add support for the `[identifier]` syntax, but it's causing a lot of conflicts, and I am not quite sure what's the best way to resolve them.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -3603,7 +3603,7 @@ module.exports = grammar({
       $._identifier,
       $._double_quote_string,
       $._tsql_parameter,
-      /`([a-zA-Z_][0-9a-zA-Z_]*)`/,
+      seq("`", $._identifier, "`"),
     ),
     _tsql_parameter: $ => seq('@', $._identifier),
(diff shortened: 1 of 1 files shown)


dmfay:
`object_id` is missing from queries/highlights.scm -- I think that's added manually and targeted to the highlight group you want for it, so `(keyword_object_id) @function.call` or similar.

This branch also seems to have gotten the last commit on main (the `select into`) under a different hash so that should be rebased out.

stackmystack:
> `object_id` is missing from queries/highlights.scm -- I think that's added manually and targeted to the highlight group you want for it, so `(keyword_object_id) @function.call` or similar.

Done
 
> This branch also seems to have gotten the last commit on main (the `select into`) under a different hash so that should be rebased out.

Done

dmfay (approved):

PR #315, opened 2025-04-09
Mssql function declaration

stackmystack:
Following up on https://github.com/DerekStride/tree-sitter-sql/pull/313.
Depends on https://github.com/DerekStride/tree-sitter-sql/pull/314.

Add support for TSQL function declaration.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -701,18 +701,21 @@ module.exports = grammar({
       ),
     ),
 
-    while_statement: $ => seq(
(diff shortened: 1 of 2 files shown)


stackmystack:
Here I tried to isolate functions, but when I tried to add WHILE, I found myself forced to change preferences all over, the commits show the evolution. So I'm not quite sure, what do you suggest?

PR #326, opened 2025-09-18
feat: support nordic/umlaut chars

### LINKED ISSUE #319: Nordic (and likely more) characters seem to break parsing of identifiers (ÅÄÖåäö)
Identifiers that use characters outside of a-Z, numbers and _ seem to break parsing.
In the example below using chars åäöÅÄÖ breaks it, also in TSQL a lot of identifiers [quotes] which also seem to break parsing, but this is in a separate issue #320 

```sql
select
    aäää = 1
    ,båå = 2
    ,cöö
    ,dÄÄ
    ,eÅÅ
    ,fÖÖ
    ,[tsql quoted expression] = 'quoted names in square brackets break'
from
    [test] t 
    inner join my_åäö_table t2 on
        t.id = t2.id
```

Image

matthias-Q:
Closes #319

Not so sure about the regex. Another option could be '/[\p{L}_][\p{L}\p{N}_]*/u' but I found that less expressive then the one we have. Otherwise we could write out the accepted chars in there, but this also does not feel right.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-latest, ubuntu-latest, windows-2019]
+        os: [macos-latest, ubuntu-latest, windows-2025]
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4


DerekStride (approved):

DerekStride:
It seems to still be stuck waiting for windows-2019, I'll force the merge and hopefully that will just resolve itself on future PRs

Issue #335, opened 2025-11-03
Add Basic Recognition for Jinja Blocks to Enable Injections


mosqueteiro:
I'd like to request adding minimal support for Jinja templating syntax to the tree-sitter-sql grammar. A similar request was opened 3 years ago (#28), however, the suggestion to use injection queries does not work because the jinja blocks break the parser making it impossible for the injection queries to hook into, much less any highlighting or indentation of the actual sql. 

I have already prepared changes to address this. They introduce 3 jinja blocks, 2 of which are just treated as extras. The goal is NOT to fully parse Jinja, but to prevent the SQL parser from breaking when encountering the jinja structures and add hook points for injection queries in templated SQL files (e.g., for dbt, airflow).

### Jinja structures:

- **Jinja Comments:** `{# ... #}`
- **Jinja Statements:** `{% ... %}`
- **Jinja Expressions:** `{{ ... }}`

I will post a PR from my branch to this repo. I'm happy to discuss the implementation details, make any requested adjustments, and ensure this change aligns with the project's standards.

mosqueteiro:
I might be misunderstanding what you are saying but to be clear I am writing sql files not embedding sql in other code. I need tree-sitter for my editor while writing sql files. So I'll have something like

**simple_example.sql:**
```sql
SELECT col1 FROM {{ my_variable_object }}
```

SQL has a very rigid structure and this parser follows that well. There is just no way to parse this without having a node the SQL parser recognizes and allows. 

In this example I might want to create a regular table in dev environment with CTAS because I need to inspect it during development but if the environment is prod it should only be a temp table because it is an intermediate step for the downstream table it feeds into

**variable_ctas.sql:**
```sql
{{ ctas_func(
    type=env_table_type,
    schema=env_schema
) }}
WITH cte1 AS (
  SELECT * FROM {{ source1 }}
)
, cte2 AS (
  SELECT
    col1,
    col2
  FROM {{ source2 }}
  WHERE col3 = some_condition
)

SELECT * 
FROM cte1
INNER JOIN cte2
  ON cte1.idx_col = cte2.col1
WHERE cte1.bool_col
```

If you parse with jinja first then inject SQL everywhere around it the SQL parser will freak out and just report ERROR because the structure is broken. Then everything that relies on tree-sitter is also broken.

DerekStride:
[...]

PR #336, opened 2025-11-03
Jinja hooks

mosqueteiro:
This adds jinja rules into the AST for the purpose of being able to hook
in injections from a jinja parser. Without these hooks the parser errors
and injections are not possible.

### Jinja structures:

- **Jinja Comments:** `{# ... #}`
- **Jinja Statements:** `{% ... %}`
- **Jinja Expressions:** `{{ ... }}`

Addresses issue: #335

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -6,6 +6,8 @@ module.exports = grammar({
     /\s/,
     $.comment,
     $.marginalia,
+    $.jinja_comment,
+    $.jinja_statement,
   ],
 
 
@@ -59,7 +61,10 @@ module.exports = grammar({
       ),
       // optionally, a single statement without a terminating ;
       optional(
-        $.statement,
+        choice(
+          $.statement,
+          $._jinja_template_statement,
+        ),
       ),
     ),
 
@@ -3434,6 +3439,7 @@ module.exports = grammar({
         $.between_expression,
         $.parenthesized_expression,
         $.object_id,
+        $.jinja_expression,
       )
     ),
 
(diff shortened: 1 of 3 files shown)


mosqueteiro:
> > So it is not SQL embedded in jinja, but jinja embedded in SQL
> 
> Jinja looks like a generic template language similar to [liquid](https://shopify.github.io/liquid/basics/introduction/), or [tree-sitter-embedded-template](https://github.com/tree-sitter/tree-sitter-embedded-template).

Yeah, it seems like it is similar to these. It uses python for rendering. The tree-sitter-embedded-template is for templates that use `<` angle brackets `>` but there are a couple jinja-specific tree-sitter repos out there.


> nvim-treesitter defines [injections on the embedded-template](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/embedded_template/injections.scm) language. Same for the [liquid injections](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/liquid/injections.scm). Unfortunately, this pattern isn't the best because you often can only partially parse the document in either parser.
> 
> I understand why this would be helpful, that said, I still don't think this belongs here.

It isn't the best with those other languages but it still works. The template blocks are either embedded in strings or, like for HTML, the language is already basically a templated text file to begin with and this other format doesn't bother the parser.

With SQL, or at least this parser, this tactic is impossible. SQL has a very rigid structure and this parser follows that well. There is just no way to parse this without having a node the SQL parser recognizes and allows.

```sql
SELECT col1 FROM {{ my_variable_object }}
```

If you parse with SQL first it just ERRORs immediately and there is nothing that a tree-sitter query can even do to locate for an injection. Also, anything relying on tree-sitter is now broken (highlights, indentation, etc.). If you parse with jinja first then inject SQL everywhere around it the SQL parser will freak out and just report ERROR because the structure is broken.

> I think that request comes from users of `dbt`, where you write SQL code with embedded jinja block.

I am using this with Airflow (data pipeline orchestrator) but the most wide-spread use of this style of SQL is with dbt, to be sure. I need templated SQL because I am deploying in different environments and need the same query to work in dev, test, and prod which are deploy on different resources. The `dbt` use-case leans much more heavily on this templating style where it needs to work across all the database and data warehouse SQL dialects. The usage of jinja in SQL is only going to continue to grow, pushed forward from the Data Analytics and Machine Learning side.

---

Btw, the `CREATE OR REPLACE TABLE` rule was missing so I added that in here too along with a test for it. If you don't want to merge my PR feel free to cherry-pick that commit out (https://github.com/mosqueteiro/tree-sitter-sql/commit/4f21a3fdd963ef43d10572c8f0333875ceda1178). I'll maintain my fork, as I need this feature supported.

PR #341, opened 2025-12-12
refactor: split large grammar.js into modules

vergenzt:
I tried to make no semantic changes to anything, but that could use some verifying. I did make substantial editorial decisions about what "belongs" where -- totally open to feedback on any of it.

I may try to do some sort of automated analysis to confirm that all of the rules expressions from before and after are the same. We'll see.

Would you be open to merging something like this?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,10 +23,11 @@ jobs:
           install-lib: false
       - run: tree-sitter generate
       - name: Run tests
-        uses: tree-sitter/parser-test-action@v2
+        uses: tree-sitter/parser-test-action@v3
         with:
           test-rust: true
           test-node: true
+          node-version: "22"
           test-python: true
           # test-go: true
           # test-swift: true
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,17 @@ endif()
 
 find_program(TREE_SITTER_CLI tree-sitter DOC "Tree-sitter CLI")
 
+add_custom_command(OUTPUT "${CMAKE_CURRENT_SOURCE_DIR}/src/grammar.json"
+                          "${CMAKE_CURRENT_SOURCE_DIR}/src/node-types.json"
+                   DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/grammar.js"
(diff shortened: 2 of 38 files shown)


vergenzt:
> I like the idea here, splitting everything out into separate modules seems nice, are you aware of other projects that make use of this pattern? We're already quite different in that we don't commit the generated code so I worry about adding more friction.

From [some searching](https://github.com/search?q=%2F%3D+require%5C%28%2F+path%3Agrammar.js+tree-sitter&type=code), the following appear to be tree-sitter grammar definitions which `require` pieces of their definitions from other files:

- https://github.com/DeNA/DeClang/blob/419618d25f9c209dc3ca3c49f72947b06f3d5fa1/mlir/utils/tree-sitter-mlir/grammar.js#L4
- https://github.com/tweakoz/ork.build/blob/71f174cdd73708268dc499c49ada4559c01dcc0b/docs/tree-sitter-cpp-grammar.js#L13
- https://github.com/quarto-dev/quarto-markdown/blob/c925e444df03c1f7b7b4cccb5f0a2e72fc130885/crates/tree-sitter-qmd/tree-sitter-markdown/old-grammar.js#L7
- https://github.com/nurulhudaapon/zx/blob/a80f92e20098a4778db839f063707d8487eb6c0d/packages/tree-sitter-zx/grammar.js#L1
- https://github.com/shackle-rs/shackle/blob/a1090067ad19e24fabc0afb751a057a1926a4b85/parsers/tree-sitter-datazinc/grammar.js#L1
- https://github.com/tree-sitter/tree-sitter-typescript/blob/75b3874edb2dc714fb1fd77a32013d0f8699989f/common/define-grammar.js#L1

To answer your question more directly though, no I'm not aware of specific other projects which split things out. I believe this is also the first major tree-sitter grammar I've attempted to make a contribution to -- but I'm pretty sure it's also the *largest* / most complex grammar I've encountered.

IMO this will substantially _reduce_ friction for new contributors because potential contributions will have a clearer place to go.

matthias-Q:
No, I am totally fine with it.

vergenzt:
Awesome! I'll try to get conflicts resolved asap and get this back to being ready to merge.

PR #353, opened 2026-02-08
Feat-jinja-support

duke8585:
## summary

adds basic jinja template support so dbt files parse cleanly instead of erroring on `{{ ref() }}`, `{% if %}`, etc.

this came up while trying to add jinja support to mergiraf - was told preprocessing is out of scope since mergiraf is language-agnostic, and the right fix is grammar-level support: https://codeberg.org/mergiraf/mergiraf/pulls/680#issuecomment-10414708

related: #335, #336

## changes

- 3 new rules: `jinja_expression`, `jinja_statement`, `jinja_comment`
- `jinja_statement` and `jinja_comment` added to extras (like whitespace/comments)
- `jinja_expression` integrated into `program`, `object_reference`, `relation`, `_expression`
- highlight queries for syntax highlighting

## test

```sh
make test
```

all 472 tests pass.

## known limitations

- no nested braces: `{{ foo }}` inside `{{ outer }}` will fail (first `}}` ends match)
- jinja inside strings is opaque: `'{{ var() }}'` parses as string literal
- no jinja validation: `{% endif %}` without `{% if %}` parses fine
- flat block structure: `{% if %}...{% endif %}` are siblings, not nested


duke8585 (commented):
related: #335, #336

this is an alternative implementation to #336 with:

simpler regex patterns (trades edge case handling for readability)
direct relation integration for cleaner FROM {{ ref(...) }} parsing
test cases focused on common dbt patterns
key differences from #336:

regex: simpler patterns that handle common cases; may need enhancement for nested braces
adds jinja_expression to relation rule (not in #336)
uses `@embedded` highlight capture for injection support

DerekStride:
Hey @duke8585, I've been thinking about this the past few weeks & I'm sticking with my original stance that this SQL parser isn't the correct place for these changes. I expand on it more here: https://github.com/DerekStride/tree-sitter-sql/issues/335#issuecomment-3969297455

PR #357, opened 2026-02-16
feat: add support for rlike and regexp

jeffbrennan:
Adds support for `RLIKE` and `REGEXP` keywords, which are used in MySQL and Spark SQL dialects. Followed the existing implementation for the `LIKE` keyword.

image


jeffbrennan:
apologies for the test failure, now passing locally 
image

PR #359, opened 2026-02-28
fix minor window function corpus test

vergenzt:
I'm designing a new language to be a superset of SQL and found this test which checks parse output for invalid SQL (which "should" be undefined).

I fixed it by fixing the syntax in the SQL query itself and running `tree-sitter test --update` to update the parse tree. It now correctly parses `FROM tab1` as a `(from ...)` node instead of another column whose expression is `FROM` and whose alias is `tab1`. 🙂


matthias-Q (approved):
Thanks!

PR #361, opened 2026-04-24
fix(scanner): don't leak start_tag across serialize/deserialize

lucieleblanc:
## Problem

I was running macOS `leaks` on a long-running host that parses many `.sql` files and noticed a growing set of allocations rooted in `tree_sitter_sql_external_scanner_deserialize` that grew linearly with the number of parses. 

I think the external scanner's `_serialize`/`_deserialize` pair mishandles the `LexerState.start_tag` for strings delimited by dollar-sign tags instead of quotes:

* `_serialize` frees and nulls `start_tag` after copying it into the snapshot buffer. If tree-sitter calls `scan()` on the same scanner instance again, it sees a NULL `start_tag` and can no longer match the pending dollar-quoted-string end tag.
* `_deserialize` assigns to `start_tag` without freeing whatever the state already owns, leaking the previous allocation every time saved state is restored.

## Solution

Free any existing `start_tag` at the top of `_deserialize` instead of `_serialize`, using the same pattern as `_destroy`.

## Testing

After this fix, running `leaks` on the same program shows no more live allocations from `tree_sitter_sql_external_scanner_deserialize`.

I added a corpus entry to assert the AST shape and guard the named-tag parse against future regressions. It produces the same AST as the previous test, but uses named tags (`$body$ ... $body$` instead of `$$`...`$$`).

### DIFF
--- a/src/scanner.c
+++ b/src/scanner.c
@@ -170,19 +170,21 @@ unsigned tree_sitter_sql_external_scanner_serialize(void *payload, char *buffer)
   }
 
   memcpy(buffer, state->start_tag, tag_length);
-  if (state->start_tag != NULL) {
-    free(state->start_tag);
-    state->start_tag = NULL;
-  }
   return tag_length;
 }
 
 void tree_sitter_sql_external_scanner_deserialize(void *payload, const char *buffer, unsigned length) {
   LexerState *state = (LexerState *)payload;
-  state->start_tag = NULL;
+  if (state->start_tag != NULL) {
+    free(state->start_tag);
+    state->start_tag = NULL;
+  }
   // A length of 1 can't exists.
   if (length > 1) {
     state->start_tag = malloc(length);
+    if (state->start_tag == NULL) {
+      return;
+    }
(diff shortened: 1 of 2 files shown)


Pyre909:
Independent confirmation, in case it helps this PR along.

LeakSanitizer, parsing through the tree-sitter 0.27.0 runtime on aarch64 Linux: **3,392 bytes in 848 allocations** from `tree_sitter_sql_external_scanner_deserialize` over roughly 4,000 parses, reached through an editor's ordinary parse path rather than a harness. The stack matches the one described here: `malloc` → `..._deserialize` (src/scanner.c:185) → `ts_parser__external_scanner_deserialize` → `ts_parser__lex`.

Worth knowing for anyone testing this: because it leaks on every state restore, it trips the leak check before anything else does, and it hides a second leak in `scan()` — the `DOLLAR_QUOTED_STRING` branch returns without freeing the tag it just scanned when that tag repeats the open one. I filed that separately as #376. Both fixes were needed before a run stayed clean.

DerekStride (approved):

PR #362, opened 2026-05-14
feat: add Firebird SQL dialect support

simeonbodurov:
## Summary

  - `program` rule now accepts `^` as an alternate statement terminator and handles Firebird's caret-mode procedures/triggers as self-terminating top-level tokens
  - New regex terminals: `set_term`, `declare_external_function`, `fb_proc_or_trigger`, `fb_var_declaration` — absorb Firebird-specific constructs atomically so complex procedure bodies
  (including triple-quote strings and `EXECUTE STATEMENT`) cannot corrupt the GLR parse state
  - Removed `^` as a binary operator (`binary_exp`) since Firebird uses it exclusively as a statement terminator
  - New keywords: `term`, `domain`, `exception`, `generator`, `active`, `inactive`, `position`, `suspend`, `exit`, `variable`, `entry_point`, `module_name`
  - New DDL rules: `create_domain`, `create_exception`, `create_generator`
  - `create_trigger` extended to support Firebird's `FOR tablename ACTIVE BEFORE/AFTER … POSITION n AS` syntax
  - `create_procedure` / create-function: added `RETURNS (params)` and `fb_var_declaration` repeat before `BEGIN`

  ## Test plan

  - [ ] All 518 existing corpus tests pass (`tree-sitter test`)
  - [ ] New `test/corpus/firebird.txt` covers `set_term`, `create_domain`, `create_exception`, `create_generator`, and Firebird trigger syntax
  - [ ] `test_speedy.py` integration test parses a 56,000-line real-world Firebird 5 schema and finds all 218 `create_table` nodes (214 regular + 4 `GLOBAL TEMPORARY TABLE`)


grammar/expressions.js
```diff
@@ -249,7 +249,7 @@ export default {
       ['*', 'binary_times'],
       ['/', 'binary_times'],
       ['%', 'binary_times'],
-      ['^', 'binary_exp'],
+      // '^' removed: Firebird uses ^ as statement terminator (SET TERM), not as power/XOR
```
DerekStride: We can't break other features

grammar/statements/create-function.js
```diff
@@ -105,6 +105,8 @@ export default {
 
   _tsql_function_body_statement: $ => seq(
```
DerekStride, line 106: Have you considered implementing a separate `_fb_function_body_statement` & branching further up the call stack?

build_and_test.py
DerekStride, line 1: Why is this here, can we remove it?

test_speedy.py
DerekStride, line 1: Can we remove this too?

DerekStride (changes requested):
There's a few definite blockers that need to be addressed before I'll do a more in-depth review.

Just my gut reaction here is that there are a lot of differences, and I'm wondering if trying to support this dialect might be more difficult than others, which might make this a challenging thing to land.

Not to say that we won't, but I think we want to clean it up a little bit to make it more isolated.

One thing that I've considered is taking a lot of these function‑body rules and moving them into a separate parser, because these procedures, functions, etc., are quite different from a SELECT statement—and an INSERT statement. 

Maybe splitting it at a DML versus DDL boundary might make sense here, and then relying on Tree‑Sitter injections to support this. That way you can load the parsers for the particular SQL dialect that you're using. We could have a generalized DML and then provide injection points for the DDL or function bodies based on a more specific parser.

But that's a long‑term play, and I don't expect that to be done here. It's just something I wanted to call out as a potential way to unblock this if it gets more complicated.

simeonbodurov:
Hi Derek,

  Thank you for the detailed review and for taking the time to look at the
PR. I appreciate the honest feedback about complexity and the long-term
vision around DML/DDL splitting with
  tree-sitter injections — that's a genuinely interesting direction.

  I should be upfront: the implementation was written by Claude Sonnet
(Anthropic's AI assistant), and I apologise in advance if that is a problem
for you. I completely understand if you
  have a policy against accepting AI-generated contributions.

  Based on your specific feedback, the following changes have been made:

  ---
  1. Restored ^ as a binary operator (grammar/expressions.js)

  ['^', 'binary_exp'] has been restored. Removing it was an overreach —
other dialects (PostgreSQL, MySQL) use ^ for XOR/power and that should not
be broken.

  To avoid the GLR conflict with Firebird's caret statement terminator: the
bare '^' has been removed from the program-rule's choice(';', '^')
terminator. This is safe because all Firebird
   caret terminators are already absorbed atomically at the lexer level —
fb_proc_or_trigger consumes CREATE...END^ as a single token, and set_term
consumes SET TERM ^ ;. No regular
  statement in Firebird scripts actually needs a bare ^ terminator in the
grammar.

  ---
  2. Separate _fb_function_body_statement
(grammar/statements/create-function.js)

  As requested, the Firebird-specific DECLARE VARIABLE declarations are no
longer mixed into _tsql_function_body_statement. A dedicated
_fb_function_body_statement rule branches from the
  same point in function_body. It uses repeat1($.fb_var_declaration) (at
least one declaration required), which naturally disambiguates it from the
T-SQL variant without needing a conflict
   declaration.

  ---
  3. Separate _fb_procedure_body_statement
(grammar/statements/create-procedure.js)

  Same approach for procedures — _tsql_procedure_body_statement is now
clean again, and a dedicated _fb_procedure_body_statement handles Firebird
procedure bodies with DECLARE VARIABLE
  declarations.

  ---
  All 518 existing corpus tests pass with no regressions.

  Thanks again for your time and the thoughtful review.

  Best regards,
  Simeon Bodurov

На чт, 14.05.2026 г. в 18:04 Derek Stride ***@***.***> написа:

DerekStride:
Hi Simeon, I have no problems with using agents to contribute, however, I don't think this PR is in a good state yet so you may want to guide it more. There's a lot of changes together making it hard to review. Could you break this up into easier to review chunks? If you're familiar with stacking PRs / using [--update-refs](https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---update-refs) it would be a good strategy for breaking this up.

simeonbodurov:
Hello Derek,

Unfortunately, I’ve only been working with Git for a month and don’t have
much experience with it. Until now, we’ve been using Visual Source Safe for
our project. If you think this PR isn’t ready for review, just ignore it.
Tree-sitter-sql with this patch works fine for me. The fork is available
for anyone who wants to use it.

На ср, 27.05.2026 г. в 18:59 Derek Stride ***@***.***> написа:

Issue #364, opened 2026-06-26
Support SET in functions


brandonchinn178:
https://www.postgresql.org/docs/current/sql-createfunction.html

This syntax should be valid:
```sql
CREATE FUNCTION public.foo()
RETURNS void
SET bar = 'asdf'
AS $$
$$;
```

PR #365, opened 2026-07-13
feat: create function with `set`

matthias-Q:
Implements #364 

(FYI: it's AI generated. I am not so familiar with the new project structure anymore. I am not a big fan of the conflict rule)


DerekStride (approved):
LGTM, conflict rule is unfortunate but we can re-evaluate if it becomes an issue

DerekStride:
Also, we can fix the windows CI in a separate PR if you want to merge this as is

PR #371, opened 2026-09-10
feat(grammar): procedural control flow in function and procedure bodies

sebsnyk:
> **First of a stack.** This PR is self-contained against `main` and is the one to review first; #372, #373 and #374 build on it in that order and are drafts until it lands. #375 is independent.

## Problem

A PL/pgSQL or PL/SQL body stops parsing at its first procedural statement. The dollar-quoted body accepts `DECLARE`, `BEGIN`, SQL statements and `RETURN <expr>`, so anything with an `IF`, a loop or an assignment produces an `ERROR` node and highlighting gives up for the rest of the file; #236 is one report of that. Run against real sources on `main`:

| source | `ERROR` nodes on `main` | with this PR |
|---|---|---|
| PostgreSQL `src/test/regress/sql/plpgsql.sql` | 939 | 636 |
| utPLSQL `source/core/ut_utils.pkb` | 292 | 233 |
| pg_partman `sql/functions/check_default.sql` | 20 | 11 |

The remainder is `RAISE`, `EXCEPTION WHEN`, `PERFORM`, `EXECUTE ... INTO`, cursors and `%TYPE`, which are separate PRs so each stays reviewable.

## Solution

A new `grammar/statements/control-flow.js` with the statements PL/pgSQL and PL/SQL share:

- assignment with `:=` or `=`, emitted as the existing `assignment` node with `left` and `right` fields
- `NULL;`
- `IF ... THEN ... ELSIF | ELSEIF ... ELSE ... END IF`
- `CASE [expr] WHEN a, b THEN ... ELSE ... END CASE`
- `LOOP`, `WHILE ... LOOP`, `FOR ... IN <int range | query | cursor> LOOP`, `FOREACH ... [SLICE n] IN ARRAY ... LOOP`, all one `loop_statement`
- `EXIT` and `CONTINUE` with an optional label and `WHEN`
- `<<label>>` on blocks and loops, and the matching `END LOOP label` / `END label`
- nested `[<<label>>] [DECLARE ...] BEGIN ... END [label]` as a `block`
- `RETURN` with no expression

They are reachable only from `function_body`, `procedure_body` and `block`, never from a top-level `statement`, so they add nothing to the states of ordinary SQL.

Things worth a second look:

- `block` and `transaction` now consume the same hidden statement list. tree-sitter deduplicates identical `repeat` helpers, and `BEGIN <statement> ;` at the top level was only unambiguous because the two rules happened to be identical. Sharing the list keeps that property; the side effect is that `transaction` tolerates procedural statements, which changes nothing for valid SQL.
- The single-statement dollar-quoted body (`AS $$ select 1 $$`) is restricted to `statement | RETURN`. With `block` as a body statement, `AS $$ BEGIN ... END $$` would otherwise be both the inline body and a single nested block.
- `1..10` lexes as the two decimals `1.` and `.10` because `_decimal_number` accepts a trailing dot. The range start is a separate integer token with lexical precedence so the dot stays with `..`. It applies only in that position; number lexing elsewhere is unchanged.
- `block` gains optional `label`, `DECLARE` and end-label children. Existing trees are unchanged.

## Notes

- `tree-sitter test`: 511 corpus tests on `main`, 524 with this PR (13 new in `test/corpus/control_flow.txt`), plus the 2 highlight tests. No `ERROR`/`MISSING` in the corpus; `scripts/test-keywords.sh` passes.
- Parse table: `STATE_COUNT` 30622 → 30898 (+0.9%), `LARGE_STATE_COUNT` 5243 → 5315, `parser.c` 41.6 MB → 42.3 MB (+1.8%). Measured with tree-sitter-cli 0.26.3 on ABI 15.
- Partially addresses #236; the sample there still needs `RAISE EXCEPTION`, coming separately.


dmfay:
hi! Before embarking on a large changeset like this and requesting significant time from the maintainers to review it, could you please talk to us directly about what you want to do and how you intend to approach it? As the issue you linked should make clear, the question of whether this grammar should cover procedural languages at all is far from settled. There are several options each with their own advantages and disadvantages, and you haven't yet made a case for your preference.

